PR #8, opened 2015-01-16
Add firewall and sysctl management modules

stumped2:
Add sysctl and firewall management modules.

Also add node definition for simple nat router to make sure the iptables rules look sane.


stumped2:
Merged

PR #16, opened 2015-10-27
Update forrest-vm puppet configuration

bdube:
* add bsd-mailx for sending mail from role account
  * forrest role account
  * ownership of apache docroot

### DIFF
--- a/data/nodes/forrest-vm.apache.org.yaml
+++ b/data/nodes/forrest-vm.apache.org.yaml
@@ -36,6 +36,6 @@ vhosts_asf::vhosts::vhosts:
     port: 80
     ssl: false
     docroot: '/var/www'
-    docroot_owner: 'forrest'
+    docroot_owner: 'forrest_role'
     docroot_group: 'www-data'
     error_log_file: 'forrest_error.log'
--- a/data/nodes/forrest-vm.apache.org.yaml
+++ b/data/nodes/forrest-vm.apache.org.yaml
@@ -6,6 +6,7 @@ classes:
   - apache::mod::proxy_http
   - apache::mod::rewrite
   - vhosts_asf::vhosts
+  - forrest_pvm_asf
 
 apache::keepalive:          'On'
 apache::keepalive_timeout:  '15'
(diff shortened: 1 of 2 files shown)


modules/forrest_pvm_asf/manifests/init.pp
```diff
@@ -0,0 +1,24 @@
...
+  package {
+    $required_packages:
+      ensure => 'present',
+  }
+
+  # manifest for forrest project vm
+
+  user { 'forrest':
+    ensure     => present,
+    name       => 'forrest',
```
stumped2: forrest is an ldap name. for conflict and making it obvious what the account is for, you should change that to forrest_role.

data/nodes/forrest-vm.apache.org.yaml
```diff
@@ -35,4 +36,6 @@ vhosts_asf::vhosts::vhosts:
     port: 80
     ssl: false
     docroot: '/var/www'
+    docroot_owner: 'forrest'
```
stumped2: docroot_owner will also need to be updated to reflect the new user name.

modules/forrest_pvm_asf/manifests/init.pp
```diff
@@ -0,0 +1,24 @@
...
+    comment    => 'forrest role account',
+    home       => '/home/forrest',
```
stumped2: This will also need to be updated with the new user name.

modules/forrest_pvm_asf/manifests/init.pp
```diff
@@ -0,0 +1,25 @@
...
+      ensure => 'present',
+  }
+
+  # manifest for forrest project vm
+
+  user { 'forrest_role':
+    ensure     => present,
+    name       => 'forrest_role',
+    comment    => 'forrest role account',
+    home       => '/home/forrest',
```
stumped2: Make sure the home path matches the username.

stumped2:
Looks good, merging.

bdube:
Cheers.

PR #17, opened 2015-11-06
inital manifest for buildbot_slave

dspt:


modules/buildbot_slaves/manifests/init.pp
```diff
@@ -0,0 +1,48 @@
+##/etc/puppet/modules/buildbot_slave/manifests/init.pp
+
+class buildbot_slave (
+
+  $system                        = true,
```
stumped2: This will actually go in the user {} and group {} sections, don't need it as a variable.

modules/buildbot_slaves/manifests/init.pp
```diff
@@ -0,0 +1,48 @@
...
+  $username                      = 'buildslave',
+
+
+# buildbot specific
+
+  user {
+    $username:
+      ensure     => $user_present,
+      name       => $username,
+      home       => "/x1/${username}",
```
stumped2: @gmcdonald Do we want the buildbot users in /x1 or do we want to put them in /home?

modules/buildbot_slaves/manifests/init.pp
```diff
@@ -0,0 +1,48 @@
...
+      groups     => $groups,
+      managehome => true,
+      require    => Group[$groupname],
+  }
+
+  group {
+    $groupname:
+      ensure => $group_present,
+      name   => $groupname,
+      gid    => $gid,
```
stumped2: This wil be dropped and replaced with ``system => true,``

modules/buildbot_slaves/manifests/init.pp
```diff
@@ -0,0 +1,48 @@
...
+# buildbot specific
+
+  user {
+    $username:
+      ensure     => $user_present,
+      name       => $username,
+      home       => "/x1/${username}",
+      shell      => $shell,
+      uid        => $uid,
+      gid        => $groupname,
```
stumped2: uid and gid will be replaced with ``system => true,``

modules/buildbot_slaves/manifests/init.pp
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  group {
+    $groupname:
+      ensure => $group_present,
+      name   => $groupname,
+      gid    => $gid,
+}
+
+  apt::source { 'utopic':
```
stumped2: As discussed in hipchat, the utopic is not needed :)

modules/buildbot_slaves/manifests/init.pp
gmcdonald: home is fine as long as the partition is large enough, what size did we do?

modules/buildbot_slaves/manifests/init.pp
gmcdonald: +1 to dropping utopic at this stage

PR #19, opened 2015-11-11
Rework buildbot_slaves module

stumped2:
Make:
* password setable
* make module more generic
* grab slave name from facter's hostname fact
* fix formatting


gmcdonald:
lgtm, will merge.

PR #21, opened 2015-12-03
Manifest for packer vms.

stumped2:
Add base manifest for packer VMs so that they have a base bootstrap.

Also update concat module because it was doing weird things at the previous version.


dspt:
lgtm :+1:

PR #25, opened 2016-01-04
get passenger up and running

rubys:
Note: puppetlabs/passenger didn't work for me, at least not with a recent version of passenger; I noted that there is a fix on master that purports to make it work, but I still couldn't get it to work.  Looking around, I found an alternative with rvm, which as an added bonus gives me a newer version of Ruby.


pctony:
Merged via CLI

PR #27, opened 2016-01-07
trigger post provisioning actions

rubys:


modules/whimsy_server/manifests/init.pp
```diff
@@ -27,12 +29,16 @@
...
+  } ->
 
   vcsrepo { '/srv/whimsy':
     ensure   => latest,
     provider => git,
     source   => 'https://github.com/apache/whimsy.git'
-  }
+  } ->
 
+  exec { 'rake::update':
```
stumped2: This exec will run *every* puppet run. 

Can you try to limit the exec to run on certain events?

exec has a creates attribute, and you can also control the running of the exec by using [audit](https://docs.puppetlabs.com/references/latest/metaparameter.html#audit) on a file.

rubys: It is intended to run on _every_ puppet run, and does nothing if there is nothing that needs to be done.  FYI: `rake` is Ruby's version of `make`.

PR #35, opened 2016-01-09
Whimsy nodejs

rubys:
install nodejs; fix puppet lint errors

### DIFF
--- a/modules/whimsy_server/manifests/init.pp
+++ b/modules/whimsy_server/manifests/init.pp
@@ -14,6 +14,7 @@ class whimsy_server (
   ############################################################
 
   $packages = [
+    build-essential,
     libgmp3-dev,
     libldap2-dev,
     libsasl2-dev,
@@ -21,6 +22,7 @@ class whimsy_server (
     zlib1g-dev,
 
     imagemagick,
+    nodejs,
     pdftk,
   ]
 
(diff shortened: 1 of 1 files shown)


modules/whimsy_server/manifests/init.pp
```diff
@@ -71,12 +79,12 @@
   #                       Mail Data Source                   #
   ############################################################
 
-  user { apmail:
-    ensure   => present,
-    uid      => 500,
+  user { 'apmail':
+    ensure => present,
+    uid    => 500,
```
stumped2: Do you absolutely need it to be uid 500? We're trying to move away from hard coded gid/uid on user creation (save for the emergency account).

If you can do `system => true` and remove the explicit uid, that'd be preferred.

rubys: I don't believe that apmail needs to be a system userid on whimsy-vm2.  I only cargo-culted the id:

    rubys@whimsy-vm:~$ id apmail
    uid=500(apmail) gid=500(apmail) groups=500(apmail),2680(sshusers)

    [apmail@hermes ~]$ id
    uid=500(apmail) gid=500(apmail) groups=500(apmail)

    apmail@minotaur:~$ id
    uid=500(apmail) gid=6(mail) groups=6(mail)

While uid of 500 may not be necessary, I'm concerned about the impact of changing this on an existing VM.  I'd be OK with tearing down the VM and recreating it from scratch.

stumped2: Changing explicit uid declaration to `system => true` shouldn't do
anything, as the user is already created. If the uid was >999 it might do
something.

PR #39, opened 2016-01-11
Run passenger apps as www-data; lock down gpg dir

rubys:


data/nodes/whimsy-vm2.apache.org.yaml
```diff
@@ -128,6 +128,9 @@ vhosts_whimsy::vhosts::vhosts:
       - /board/publish_minutes
 
     custom_fragment: |
+      PassengerDefaultUser www-data
+      PassengerDefaultGroup www-data
```
stumped2: If possible, hard coding `www-data` is bad. There should be a way to pull this user and group from the [puppet-apache](https://github.com/puppetlabs/puppetlabs-apache/tree/1.6.0) module.

PR #73, opened 2016-03-31
Modify taverna membership after mentor swap

stain:
See http://incubator.apache.org/projects/taverna.html

> Chris Mattman ➝ Dan Debrunner


gmcdonald:
thanks for the pr, taking a look

PR #77, opened 2016-04-22
Redirects to home.apache.org

stain:
Strawman (untested) 

HTTP permanent redirect from http://people.apache.org/* to https://home.apache.org/*  -- ensures a single hostname

Note how the "people" vhosts are the `default_vhost` so they redirect for any other hostnames as well. - you may want to change this as it can complicates testing on other hosts.

However this change does not enforce https:// as there (should not be) any sensitive data on home.


gmcdonald:
Sorry, its just way out of date now (our fault.)

We.d need a new PR with fully tested code.

PR #105, opened 2016-07-17
grant secretary write access to hudson-jobadmin

rubys:


gmcdonald:
We have worked around this by making Sam a hudson-admin

PR #112, opened 2016-07-30
rough in group self service proposal

rubys:
See the following thread for background:

https://lists.apache.org/list.html?infrastructure@apache.org:lte=1M:make%20groups%20self%20service

Details:

1) add owner to list of hudson apps attrs, and grant owners write access to
   hudson apps.  (leaving hudson-admin and pmc-chairs in place for now).

2) add ou=auth permissions, modelled after parent (ou=groups) with the following
   changes: members of the group have write permissions, and pmc-chairs
   are not listed.

3) add owner to list of ou=groups attributes, and grand owners write access
   to groups.  (leaving pmc-chairs in place for now).

4) change db-jdo definition in asf-authorization from an inline list to
   a reference to LDAP.


rubys:
Committed: https://github.com/apache/infrastructure-puppet/commit/f76aad5ab67480414c67136120cd9bfd01cb9b56 and https://github.com/apache/infrastructure-puppet/commit/1b32a70baab832a64c806ed991ed7a05db3fe792

PR #113, opened 2016-07-31
convert %(...) => %{...} in vhost custom_fragments

rubys:
Background: hiera and Apache unfortunately have the same escape sequence for
interpolating variables in strings.  While there are various workarounds for
escaping the escape sequence (%%{}{var} and %{literal('%'}}{var} are both
popular), neither are working for me when trying to do <If> statements.

After more than an hour of trying various alternatives, it seems best to
avoid the syntax clash altogether.  As this change will need to be deployed
and the puppet master restarted before it can be used, I'm pushing it
separately.


rubys:
Committed: https://github.com/apache/infrastructure-puppet/commit/0e8616ff7b31356a40749c51c8ee0933ff43f823

PR #116, opened 2016-08-02
convert non-podling inline lists to LDAP auth groups

rubys:


johnament:
Does this break phonebook?

Humbedooh:
Phonebook should be fine with this :)

PR #130, opened 2016-10-21
INFRA-12291 remove git package from gump_server, already added via gi…

bodewig:
…t_client

I just realized puppet agent wasn't picking up my changes as the git packages was added by both `gump_server` and `git_asf::git_client`. TBH I'm not sure how `git_asf::git_client` is part of the classes that apply to `vmgump3`.


bodewig:
via `ubuntu/14.04`, I see.

bodewig:
there'll likely be more conflicts

PR #137, opened 2016-10-26
INFRA-12291 first attempt at configuring gump's httpd

bodewig:


gmcdonald:
I see you removed ant and tomcat also by overriding in the yaml the base packages. That is fine, but that should have been a different commit (same push is fine) with a commit message that explains this change. iow, it has nothing to do with the setting up of httpd.

bodewig:
I've split it (and rebased the branch again), thought you'd prefer a squashed commit, sorry. :-)

gmcdonald:
merged, looks to be working well

bodewig:
Yes, looks good, I may need to take core of https in a way (HTTPS everywhere sends me there and it looks broken). I'll need a few more steps but hope to finish the migration within the next week or so.

Thanks.

PR #139, opened 2016-10-29
INFRA-12291 Vmgump3 setup

bodewig:
This contains several steps necessary to replace vmgump. If all works the way I expect it to the only issue remaining is that vmgump-vm3 must be able to send emails -> #140

### DIFF
--- a/modules/gump_server/manifests/cronjobs.pp
+++ b/modules/gump_server/manifests/cronjobs.pp
@@ -33,4 +33,12 @@ class gump_server::cronjobs {
     hour => [6,12,18],
     minute => '0',
   }
+
+  cron { 'Official Gump run sending out nag mails':
+    ensure => present,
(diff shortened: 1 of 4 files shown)


modules/gump_server/manifests/cronjobs.pp
```diff
@@ -17,4 +17,28 @@
     hour => '0',
     minute => '0',
   }
+
+  cron { 'Clean up logs older than about a month':
```
stumped2: You could combine these into a single stanza:

```ruby
cron {
  'Clean up...':
    ensure => present,
    ...
    minute => 0;
  'Public - ...':
    ensure => present,
    ...
    minute => 0;
}
```

Also, please make sure to stay consistent with you formatting, as numbers should not be quoted.

Please see https://github.com/stumped2/puppet-hooks/blob/master/hooks/pre-commit for puppet-lint styling.

PR #140, opened 2016-10-29
INFRA-12291 allow relaying of emails from vmgump-vm3

bodewig:


gstein:
Is it possible to have comments in that yaml file? If so, then I'd like to see the hostname recorded in a comment for that IP.

(and Infra should backfill those other networks/IPs)

bodewig:
I've added the IPs I could easily resolve.

gstein:
Cool! Thanks!

PR #142, opened 2016-10-30
INFRA-12291 enable https for vmgump-vm3.apache.org

bodewig:
This is not strictly necessary to resolve INFRA-12291 but people using "HTTPS Everywhere" currently can't access vmgump-vm3.


Humbedooh:
I don't think we'd want to install a wildcard cert on that VM, that seems quite dangerous.
Why not use Lets Encrypt instead?

bodewig:
I agree, just mimiced what I saw on other machines. I'll try to figure out how to use Lets Encrypt.

PR #146, opened 2016-11-04
Updating reporeq to not prompt for github integrations.

johnament:


johnament:
screen shot 2016-11-03 at 9 31 15 pm

PR #147, opened 2016-11-07
puppet-lint gump_server's manifests

bodewig:
based on review comments by @stumped2 on #139


gmcdonald:
There is more going on here than just a lint check.
You have broken out some code into its own class (and file); however I do not see a reference to that class anywhere (such as the yaml) so it wont get used.
Can you confirm your intentions re the new pp file in this patch

bodewig:
`gump_server::opt_package` is used from within `init.pp`.

puppet-lint complained about me defining new resource types inside a class, so I moved it out.

stumped2:
What @gmcdonald said. You moved it into a new class, but it is not being referenced anywhere. Somewhere in the rest of the ``gump_server`` module, you have to pull in the ``opt_package`` define, otherwise it will not be used at all.

bodewig:
Isn't it referenced in https://github.com/apache/infrastructure-puppet/pull/147/files#diff-f68ac488cddbe4be9f000ccbb31cc453R107 ff? It's my understanding puppet will find it when used by name (and it works in my local vagrant testbed :-).

gmcdonald:
ok, didnt see those, I was expecting either an include or the class being pulled in from the yaml file.

Will merge shortly.

gmcdonald:
All done

bodewig:
Thanks!

PR #156, opened 2016-12-18
First pass of the configuration for the Syncope PoC VM node

PierreSmits:
Please review the changes regarding the configuration of the sycope-vm2 node for the Syncope PoC, see https://issues.apache.org/jira/browse/INFRA-10931


PierreSmits:
Thank you for the fast reaction.

I will look at the suggestion and rework the syncope-vm2 node configuration accordingly.

A small question though: what is it that needs the JIRA ticket for the DNS? 
^ I am quite new to doing this kind of stuff, and your 'submit a jira ticket for the DNS' is a bit vague to me.

Best regards,

Pierre

clambertus:
Your PR contains a (commented) reference to idm-poc.apache.org. I was going to have you create a ticket for infra to create it, but I was in the zone files doing something else, so I already took care of adding idm-poc.a.o DNS. When you set up the letsencrypt yaml block, add idm-poc as another cert domain so the certificate will encompass both syncope-vm2.a.o and idm-poc.a.o.

PierreSmits:
Done. Suggested changes incorporated.

data/nodes/syncope-vm2.apache.org.yaml
```diff
@@ -27,6 +27,21 @@ apache::mod::event::startservers: '5'
 apache::mod::event::threadlimit: '500'
 apache::mod::event::threadsperchild: '50'
 
+apache::mod::ssl::ssl_cipher: 'HIGH:MEDIUM:!aNULL:!MD5:!RC4:!DHE-RSA-AES128-SHA'
+
+letsencrypt::email: 'root@apache.org'
+letsencrypt::manage_dependencies: false
+letsencrypt::certonly:
+  security-vm:
```
clambertus: This needs to be 'syncope-vm2'

data/nodes/syncope-vm2.apache.org.yaml
```diff
@@ -27,6 +27,21 @@ apache::mod::event::startservers: '5'
...
+    domains:
+      - idm-poc.apache.org
+      - syncope-vm2.apache.org
```
clambertus: Please switch the order here, the first domain in the domain list is what the certificate file is named, and needs to reflect the path specified in the ssl_* paths, e.g. ssl_cert: '/etc/letsencrypt/live/syncope-vm2.apache.org/cert.pem'

data/nodes/syncope-vm2.apache.org.yaml
```diff
@@ -1,3 +1,97 @@
...
+oraclejava::install::version: '8'
+oraclejava::install::unlimited_jce: true
+
+# idm_poc class variables.
+# idm_poc::syncope_version: '2.0'
+# idm_poc::syncope_revision_number: '1'
+# idm_poc::mysql_connector_version: '5.1.11' <- this must switch to postgresql client
+# idm_poc::parent_dir: '/usr/local/syncope' <- this must be verified
+# idm_poc::data_dir: '/usr/local/syncope_data' <- this must be verified
+# idm_poc::connector_port: '9080' <- the default port for syncopy
```
ilgrosso: it's "Syncope", not syncopy

And 9080 is not the default port at all.

ilgrosso: The actual port simply depends on Java EE container configuration; 9080 is the port of the Standalone distribution.

Here we are not going to use such distribution, but, as said, the Maven project built from https://git1-us-west.apache.org/repos/asf?p=iampoc.git

So the process to implement (possibly here, I am not familiar with puppet) is to:

1. prepare a Tomcat 8.5 container with at least DataSource configuration for local PostgreSQL
1. git clone / git pull from the repo above
1. build as reported in the [reference guide](https://ci.apache.org/projects/syncope/reference-guide.html#customization)
1. copy the 3 target WAR files into the configured Tomcat container
1. start Tomcat

data/nodes/syncope-vm2.apache.org.yaml
```diff
@@ -1,3 +1,97 @@
...
+    serveraliases:
+      - 'idm-poc.apache.org'  # this must be set up
+    port: 443
+    docroot: '/var/www'
+    error_log_file: 'idm_error.log'
+    ssl: true
+    ssl_cert: '/etc/letsencrypt/live/syncope-vm2.apache.org/cert.pem'
+    ssl_key:  '/etc/letsencrypt/live/syncope-vm2.apache.org/privkey.pem'
+    ssl_chain: '/etc/letsencrypt/live/syncope-vm2.apache.org/chain.pem'
+    custom_fragment: |
```
ilgrosso: The working reverse proxy configuration for is like as https://paste.apache.org/6WYU

data/nodes/syncope-vm2.apache.org.yaml
```diff
@@ -1,3 +1,97 @@
...
 classes:
+  - apache
+  - apache::mod::proxy
+  - apache::mod::proxy_http
+  - apache::mod::rewrite
+  - letsencrypt
   - loggy
+  - maven::maven    # can't find any reference regarding maven/mvn
+  - oraclejava::install
+  - postgresql::client
```
ilgrosso: We need the actual PostgreSQL DBMS: is it `postgresql::server`? Which version?

data/nodes/syncope-vm2.apache.org.yaml
PierreSmits: Thank you for the review, Francesco.

As you can see all the lines the 'idm-poc' segment is commented out. As far as I can tell at the moment there is no puppet component/module available, and it might even be so that there never will be. I just included this segment as a set of configuration hints/references for myself, and potential others who might feel the itch to scratch that puppet module itch.

I can remove it when desired.

data/nodes/syncope-vm2.apache.org.yaml
PierreSmits: I rather see it happening that we utilise the existing PostGreSQL of the ASF, if that is possible. It alleviates the burden on us (with respect to setting up and managing yet another environment). In that case we can suffice with the client. 

Infra should be able to say whether that is possible, and if that is not their call they should be able to point us in the right direction.

data/nodes/syncope-vm2.apache.org.yaml
PierreSmits: Thank you, Francesco, for providing some additional insight regarding the proxy configuration. 
I will adjust the custom_fragment in the vhosts segment accordingly. Just need to figure out how to do such in relation to the puppet configuration.

data/nodes/syncope-vm2.apache.org.yaml
ilgrosso: I can understand your position, let's see how it will go.
Personally, I believe this VM needs a dedicated DBMS (because it is a self-contained PoC, with an embedded LDAP server too).

data/nodes/syncope-vm2.apache.org.yaml
PierreSmits: Done!

PR #173, opened 2017-01-25
Adding mxnet entries

hen:


johnament:
@hen is mxnet using SVN? If not, then this isn't needed.

hen:
Ah - didn't realize. It's using GitHub. I was following the 'Authorize Committers' section of the Mentors page and http://www.apache.org/dev/infra-contact#requesting-podling.

I see that permissions for GitHub are Incubator wide; so do I just follow the http://www.apache.org/dev/pmc.html#karma instructions for each committer?

PR #174, opened 2017-01-31
Add ratis to auth template

jnp:
Please review the changes to asf-authorization-template to add new podling ratis.


johnament:
Similar to the question asked for MXNet.  Is Ratis using SVN for source control? If not, this section is not needed

```
+[/incubator/ratis]
 +@ratis = rw
```

jnp:
Thanks John. I have removed this section.

PR #181, opened 2017-02-16
We don't need SSL and certificate for videos!

JacquesLeRoux:
Trying to fix the certificate issue we have with OFBiz demos

### DIFF
--- a/data/nodes/ofbiz-vm2.apache.org.yaml
+++ b/data/nodes/ofbiz-vm2.apache.org.yaml
@@ -164,20 +164,3 @@ vhosts_asf::vhosts::vhosts:
     docroot_group: 'www-data'
     error_log_file: 'ofbiz-bigfiles_error.log'
     custom_fragment: |
-      RedirectMatch permanent ^/(.*)$ https://ofbiz-vm2.apache.org/$1
-
-  ofbiz-vm-bigfiles-443:
-    vhost_name: 'ofbiz-vm2.apache.org'
-    servername: 'ofbiz-vm2.apache.org'
-    serveraliases:
-      - 'www.ofbiz-vm2.apache.org'
-    port: 443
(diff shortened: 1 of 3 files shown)


PierreSmits:
This is not going to fly when you leave 'RedirectMatch permanent ^/(.*)$ https://ofbiz-vm2.apache.org/$1' in place.

JacquesLeRoux:
Right Pierre, it's indeed useless now. I'll fix that!

PierreSmits:
It was working, right?

So now we don't have a secure path to ofbiz-vm2.apache.org, a letsencrypt config that is not going to be used...

JacquesLeRoux:
The whole was not working as reported at OFBIZ-9206 and INFRA-11960. So I try to make the setup as simple as possible. For instance we don't need a certificate to shows static files (videos and such). If you have better ideas please share...

PierreSmits:
It is working. See https://ofbiz-vm2.apache.org/FrameworkIntro/

I wonder why we would go through all this trouble to correct a working setup. I recommend to leave it as is.

PierreSmits:
The only part that is not working is the backend OFBiz demos (see earlier mentioned JIRA issues), which is due to OFBiz demo specific setups. The letsencrypt and httpd configs work as intended,  see:
https://demo-old.ofbiz.apache.org/catalog/control/main
https://demo-stable.ofbiz.apache.org/catalog/control/main
https://demo-trunk.ofbiz.apache.org/catalog/control/main

JacquesLeRoux:
Right, I also tried from there https://cwiki.apache.org/confluence/display/OFBIZ/Framework+Introduction+Videos+and+Diagrams. I'll remove my change it's useless.

The pb is indeed that OFBiz is adding the port 8443. I'll check why the demos are doing that.

PR #191, opened 2017-04-09
support multiple LDAP groups via requireany

rubys:


rubys:
Since this pull request contains macro changes: merge #191 and then once picked up, restart the puppet server

PR #203, opened 2017-05-08
Git homepage update

Gagravarr:
Mention Gitbox, expand the writable git details a little bit, and update infra list to direct queries to


pono:
Thanks!

PR #205, opened 2017-05-24
allow LDAP attribute to be specified in authorization templates

rubys:
look for ";attr=" within {ldap:} sections, and extract attribute name
if this string is present; otherwise default to memberUid,member for
backwards compatibility.

A single example of this new syntax is included in this commit.

The plans are to shortly make use of this to move PMC authorizations to
an owner attribute.

### DIFF
--- a/modules/subversion_server/files/scripts/authorization/gen_asf-authorization.pl
+++ b/modules/subversion_server/files/scripts/authorization/gen_asf-authorization.pl
@@ -93,7 +93,7 @@ sub rebuild {
             die "Uh-oh: found '@lines', expected one line; at '$_'" unless @lines == 1;
             $_ = $lines[0];
         }
-        if ( $_ !~ m/^#/ && $_ =~ m/{ldap:(cn=[^;]*);?(.*)}/ ) {
+        if ( $_ !~ m/^#/ && $_ =~ m/{ldap:(cn=[^;\s}]*);?([^\s}]*)}/ ) {
             my ( $dn, $opts ) = ( $1, $2 );
             chomp;
             my @groupdn = split( /,/, $dn );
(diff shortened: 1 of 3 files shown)


modules/subversion_server/files/scripts/authorization/gen_asf-authorization.pl
```diff
@@ -93,15 +93,17 @@ sub rebuild {
             die "Uh-oh: found '@lines', expected one line; at '$_'" unless @lines == 1;
             $_ = $lines[0];
         }
-        if ($_ !~ m/^#/ && $_ =~ m/{ldap:(cn=.+)}/ ) {
+        if ( $_ !~ m/^#/ && $_ =~ m/{ldap:(cn=.+?)(?:;(.*)|)}/ ) {
```
sebbASF: Might be better to use something like:

m/{ldap:(cn=[^;]+)(?:;(.*))}/

This will avoid any potential backtracking due to the '?' qualifier

sebbASF: Forgot to make the attribute optional, should be (untested)

m/{ldap:(cn=[^;]+)(?:;(.*))?}/

rubys: Optimized further: https://github.com/apache/infrastructure-puppet/pull/205/commits/76ed3ddf4f5ff0ed33608989485680787931da11

sebbASF: Neat!

rubys:
Related discussion: https://lists.apache.org/list.html?private@infra.apache.org:lte=10M:Setting%20up%20podlings%20in%20LDAP

PR #207, opened 2017-05-28
allow ASF members to update group ACLs

rubys:
Hold on merging this until I say so.  I specifically want to get the following, in order: (1) the OK of David and/or Greg, (2) to pass it by the board, and then (3) pass it by the membership.

Discussion thread:

https://lists.apache.org/thread.html/9f79cce1ab235cd916128988a67303941d22ae9f42e2ac3eba1f7483@%3Cprivate.infra.apache.org%3E

LDAP documentation:

http://www.openldap.org/doc/admin24/access-control.html#Group%20ACLs%20without%20DN%20syntax

### DIFF
--- a/modules/ldapserver/templates/slapd.conf.erb
+++ b/modules/ldapserver/templates/slapd.conf.erb
@@ -261,6 +261,7 @@ access to dn.subtree="ou=groups,dc=apache,dc=org"
   by group.exact="cn=apldap,ou=groups,ou=services,dc=apache,dc=org" write
   by group.exact="cn=asf-secretary,ou=groups,ou=services,dc=apache,dc=org" write
   by group.exact="cn=pmc-chairs,ou=groups,ou=services,dc=apache,dc=org" write
+  by set="[cn=member,ou=groups,dc=apache,dc=org]/memberUid & user/uid" write
   by dnattr=owner write
   by dn="cn=ldaprepl,ou=users,ou=services,dc=apache,dc=org" read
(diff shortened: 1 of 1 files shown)


modules/ldapserver/templates/slapd.conf.erb
```diff
@@ -272,6 +272,7 @@ access to dn.subtree="ou=groups,dc=apache,dc=org"
 access to dn.exact="cn=pmc-chairs,ou=groups,ou=services,dc=apache,dc=org"
```
gstein: Maybe I'm missing something, but the above seems to imply Members can change the set of PMC *chairs* rather than the membership of the PMCs. ??

modules/ldapserver/templates/slapd.conf.erb
```diff
@@ -272,6 +272,7 @@ access to dn.subtree="ou=groups,dc=apache,dc=org"
...
   by group.exact="cn=apldap,ou=groups,ou=services,dc=apache,dc=org" write
   by group.exact="cn=asf-secretary,ou=groups,ou=services,dc=apache,dc=org" write
+  by set="[cn=member,ou=groups,dc=apache,dc=org]/memberUid & user/uid" write
```
gstein: After looking at the whole file, I believe this line needs to move about 10 lines upwards.

gstein:
Infrastructure is +1 for this change. It assists with the deprecation of the LDAP scripts on minotaur.

PR #208, opened 2017-05-30
INFRA-13336 Adding jbake as a package

johnament:


gmcdonald:
Thanks, there is more to do. Grep the same file you edited for 'forrest'  - you will see more places that jbake needs to be mentioned.

Please also see the data/ubuntu/[14|16]04.pp files those need an entry too

PR #216, opened 2017-06-18
INFRA-13213 focus on metadata only: *.pom maven-metatada.xml *.sha1 *.asc

hboutemy:
this will reduce drastically the volume of mirrored data
but will contain main information, particularly what proves central immutability


gstein:
iirc, this VM is turned off, so we should just close this PR?

hboutemy:
sure, too late :/

PR #219, opened 2017-06-26
INFRA-14329: Add auth for openwhisk ppmc

Humbedooh:
As subject says, this will add auth for the openwhisk PPMC. This is probably the first time we try this with the new podling LDAP structure, but it should just do the right thing.


rubys:
LGTM.  It also incidentally is a test of the change I made where the name to the left of the equals can be anything other than the groupdn with -pmc optionally appended to it.

https://github.com/apache/infrastructure-puppet/commit/a67bed164e7ad45e60f8109757654d17ec82ef78

PR #228, opened 2017-06-29
allow members + PMC chairs to issue account requests

rubys:


sebbASF:
The patch removes @exec-officers = rw - is that intentional?
It's not detailed in the commit message

rubys:
I believe that exec-officers is a proper subset of members.  Also, it doesn't appear that exec-officers has been kept up to date.

clambertus:
Agreed regarding exec-officers.

PR #229, opened 2017-06-29
enable whimsy to submit mlreqs on behalf of committers

rubys:


rubys:
This is needed to implement a fix for https://github.com/apache/whimsy/issues/35

PR #233, opened 2017-07-05
Christ/add mergebot role to gitbox

cthistle:


cthistle:
+1 via humbedooh in hipchat

PR #244, opened 2017-07-17
allow ASF members and members of hudson-jobadmin to update hutson-jobadmin

rubys:
I note that the following was stated in https://issues.apache.org/jira/browse/INFRA-14616:

> A PMC member should be able to add you to the hudson-jobadmin group via
> https://whimsy.apache.org/roster/group/hudson-jobadmin

While this is the intended direction, we aren't there yet.  For two reasons:

* I don't know how to say "any member of any PMC" in slapd.conf.
* The way that PMCs are being stored in LDAP is changing.

What is straightforward to do right now is to allow any member of hudson-jobadmin to update hudson-jobadmin.  And to allow any ASF member to do so.

This patch does that.

If this patch is applied and deployed, I'll update the whimsy roster tool.


johnament:
I'm -1 to this change.  I add random podling developers to this, and that means they'd be able to add more.

gmcdonald:
I too am -1 for hudson-jobadmin to add other hudson-jobadmin 

Either make it PMC Members, or if you cant work that out, leave it as-is , i.e. PMC Chairs only, thats fine.

gstein:
PMC Chairs + ASF Members ?

gmcdonald:
As per gstein comment, yes please, can you alter this to be PMC Chairs + ASF Members.
If not then PMC Chairs only is fine.

gmcdonald:
ok, do a new PR when you get back to this then.

PR #261, opened 2017-07-28
Added RequestHeader unset Accept-Encoding to the vhost for donate.a.o

cthistle:


cthistle:
Got +1 from @Humbedooh in hipchat.

PR #265, opened 2017-08-02
Update kif.py

gstein:
Trim the noisy hipchat messages.


Humbedooh:
This will be a break from upstream, but I'm sure we can manage that :) lgtm

PR #266, opened 2017-08-02
Fix up redirects to be relative (and correct wrong redirect in vm3!)

sebbASF:


gstein:
Seems these redirects should be marked as permanent redirects.

sebbASF:
Possibly, but IMO that is a separate fix

gstein:
If you are changing these redirects, then *also* make them permanent.

PR #269, opened 2017-08-02
Update .htaccess ; add .sha512

henkp-apache:


sebbASF:
The commit message does not agree with the change, which adds .sha128, not .sha512

I assume it was supposed to add both?

PR #272, opened 2017-08-04
Redirect to current so still works when transitioning to new instance

sebbASF:
Whilst the existing redirect works, it would need to be changed if a new instance is created and made current.


rubys:
:+1:

PR #273, opened 2017-08-05
Allow hermes apmail login to check SVN sync status

sebbASF:
This is needed for INFRA-14792.


sebbASF:
It would be helpful if this could be deployed please.

The apezmlm login on hermes already has access to the files since they are checked out locally.
The change just allows it to check that they are up to date

PR #275, opened 2017-08-06
fix LDAP lookups for reporeq

Humbedooh:
Use HTTP request (for now?) tro fetch podlings, filter them out of LDAP.
Also adjust LDAP to new structure and closest server.

### DIFF
--- a/modules/git_self_serve/templates/ss.lua.erb
+++ b/modules/git_self_serve/templates/ss.lua.erb
@@ -1,6 +1,6 @@
 local JSON = require 'cjson'
 local http = require 'ssl.https'
-local ldap_host = "snappy5.apache.org"
+local ldap_host = "ldap1-lw-eu.apache.org"
 local aaa_domain = "apache.org"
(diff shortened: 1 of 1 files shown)


sebbASF:
This change looks wrong.
It only allows for PMCs to be defined using ou=project (new LDAP) and not ou=committees (old LDAP).

Hardly any PMCs use the new LDAP scheme yet.

So both queries are needed.

rubys:
All PMCs are defined using the new LDAP scheme.  Most PMCs also are defined using the old LDAP scheme.  The whimsy roster tool (which now should be the only way that committee rosters are updated) will update both lists  for non "guineapigs".

I had hoped by now to have deleted all of the old LDAP definitions, but we continue to encounter tools that use it; but it should be safe for all tools to move exclusively to depending on the new LDAP definitions to be in place.

sebbASF:
OK, I see.

I must have missed that change while I was away recently.

PR #280, opened 2017-08-09
Fetch projects.json

sebbASF:
home.a.o needs to use ldap_projects.json

### DIFF
--- a/data/nodes/home.apache.org.yaml
+++ b/data/nodes/home.apache.org.yaml
@@ -80,6 +80,11 @@ cron:
     minute: '9'
     hour: '*'
     command: 'cd /var/www/html/public/ && wget -N -nv https://whimsy.apache.org/public/public_ldap_authgroups.json'
+  json10:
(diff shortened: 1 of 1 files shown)


pono:
I'm happy to merge this, but curious why it needs this file.

sebbASF:
home.a.o provides info on committees and podlings.

These are now in projects.json which will eventually take over from groups, committees and nonldap_groups.

clambertus:
Are you referring to home.a.o/phonebook? isn't this almost entirely superseded by whimsy's roster tool at this point?

sebbASF:
Whilst Whimsy provides much the same functionality as phonebook, it requires a committer login.
Phonebook is much faster, though it is not as up to date.

Also home.a.o creates the keys files which Whimsy does not.
That is what needs the extra file copy.

However if someone wants to take the decision to retire the home.a.o functions, I can look into providing redirects to the relevant Whimsy URLs.

PR #281, opened 2017-08-09
move svn auth to use ou=project,ou=groups,dc=apache,dc=org

rubys:
This should make ou=pmc,ou=committees,ou=groups,dc=apache,dc=org and ou=groups,dc=apache,dc=org redundant for non "guinea pig" PMCs.


rubys:
I tested it locally, and found zero differences in the output produced before and after this patch is applied.

PR #294, opened 2017-08-11
Drop obsolete copy of gen_asf-authorization.pl

sebbASF:
The up-to-date copy is at:
modules/subversion_server/files/scripts/authorization/gen_asf-authorization.pl


cthistle:
This file is referenced in the init.pp line [571](https://github.com/apache/infrastructure-puppet/blob/0a39ffe1d96e134586f553950677f10d3c4a67d1/modules/subversion_server/manifests/init.pp#L571) and [580](https://github.com/apache/infrastructure-puppet/blob/0a39ffe1d96e134586f553950677f10d3c4a67d1/modules/subversion_server/manifests/init.pp#L580)

sebbASF:
In which case the references need to be updated, because the file is considerably out of date with the one in

modules/subversion_server/files/scripts/authorization

Are you sure that one in init.pp is derived from the obsolete copy?
I could not work out where /x1/svn/scripts is sourced.

cthistle:
My mistake, those lines are calling ../modules/subversion_server/files/scripts/authorization so that's all good. In that case, I have no idea why there is a separate ../modules/subversion_server/files/authorization/ copy.

sebbASF:
The scripts/ version was added on Dec 2014 and has been updated regularly.

The obsolete copy was also added in Dec 2014 a few minutes later but was never updated.
Looks like it was a mistake.

PR #302, opened 2017-08-12
Replace/remove local group definitions

sebbASF:
There are now no local podling or pmc group definitions left in the auth files.


gmcdonald:
The commit for ' Zest => Polygene;' removes Zest but does not insert Polygene auth, is that what was intended?

sebbASF:
Yes, that was deliberate.

PR #311, opened 2017-08-16
from henkp - add support for sha512

pono:


pono:
obvious fixing after approval from henkp

PR #321, opened 2017-08-22
Jenkins Node Monitor

cthistle:
New module that pulls node and status from builds.a.o and pushes that data to DataDog as a custom service check.

### DIFF
--- a/modules/jenkins_node_monitor/files/jnm.py
+++ b/modules/jenkins_node_monitor/files/jnm.py
@@ -4,8 +4,7 @@
 import ConfigParser
 from datadog import initialize, api
 from datadog.api.constants import CheckStatus
-import json
-import urllib
+import requests
 
 config = ConfigParser.ConfigParser()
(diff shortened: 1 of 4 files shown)


modules/jenkins_node_monitor/files/jnm.py
```diff
@@ -0,0 +1,28 @@
...
+import urllib
+
+config = ConfigParser.ConfigParser()
+config.read("settings.cfg")
+options = {'api_key': config.get("datadog_agent", "api_key")}
+initialize(**options)
+
+url = "https://builds.apache.org/computer/api/json"
+check = 'jenkinsNode.status'
+response = urllib.urlopen(url)
```
Humbedooh: why not simplify this:

~~~
   import requests
   jenkinsNodes = requests.get(url).json()
~~~

Humbedooh: and then ditch both urllib and json imports

PR #322, opened 2017-08-23
tweak id.a.o design to conform to new self serve portal.

Humbedooh:
tweak id.a.o design to conform to new self serve portal.


Humbedooh:
No need to merge this just yet, it's just here 'cause..

PR #331, opened 2017-08-23
add in some suggestions from sebb

Humbedooh:
add in some suggestions from sebb

### DIFF
--- a/modules/selfserve_portal/files/site/cgi-bin/createml.cgi
+++ b/modules/selfserve_portal/files/site/cgi-bin/createml.cgi
@@ -25,7 +25,6 @@ import requests
 import base64
 import subprocess
 import re
-import uuid
 import sscommon
 
 requser = os.environ['REMOTE_USER']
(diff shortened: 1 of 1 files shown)


modules/selfserve_portal/files/site/cgi-bin/createml.cgi
```diff
@@ -89,10 +88,7 @@ payload = {
     'mods': mods
 }
 
-uid = uuid.uuid4()
-with (open("/usr/local/etc/selfserve/queue/%s-%s-%s.json" % (uid, listname, domain), "w")) as f:
-    json.dump(payload, f)
-    f.close()
+json.dump(payload, open("/usr/local/etc/selfserve/queue/mailinglist-%s-%s.json" % (listname, domain), "w"))
```
sebbASF: Only the f.close() should be deleted.
The with() block is still needed to establish the context for the automatic close.

PR #333, opened 2017-08-23
INFRA-15000: change users@ to private@ - less spam for ordinary folks

Humbedooh:
change users@ to private@ - less spam for ordinary folks


Humbedooh:
gnomes for pono

Humbedooh:
more gnomes

PR #378, opened 2017-08-30
let's ditch the private checkbox for now.

Humbedooh:
let's ditch the private checkbox for now.

### DIFF
--- a/modules/selfserve_portal/files/site/mail.html
+++ b/modules/selfserve_portal/files/site/mail.html
@@ -25,8 +25,8 @@
           please file a JIRA ticket with Infra.
           </p>
           <p>
-          You may use the shortcut <b><kbd>podlists</kbd></b> as a new list name, which will
-          create both private@, dev@ and commits@ in one go.
+          Any <b><kbd>private@</kbd></b> and <b><kbd>security@</kbd></b> lists created will always be private.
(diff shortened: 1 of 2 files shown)


modules/selfserve_portal/files/site/mail.html
```diff
@@ -25,8 +25,8 @@
           please file a JIRA ticket with Infra.
           </p>
           <p>
-          You may use the shortcut <b><kbd>podlists</kbd></b> as a new list name, which will
-          create both private@, dev@ and commits@ in one go.
+          Any <b><kbd>private@</kbd></b> and <b><kbd>security@</kbd></b> lists created will always be private.
```
sebbASF: Is the podlists shortcut no longer valid?

modules/selfserve_portal/files/site/mail.html
```diff
@@ -25,8 +25,8 @@
...
+          Other lists that are created as private should go through a JIRA ticket.
           </p>
```
sebbASF: Does not read well. Perhaps put:

Other lists that are to be created as private must be requested in a JIRA ticket.

modules/selfserve_portal/files/site/mail.html
Humbedooh: no, this was replaced by a predefined list to pick from

PR #399, opened 2017-09-05
Don't use Received headers if List-Post is present

sebbASF:
At present if the List-Post is present but does not reference an apache.org domain, the code will try to use the Received headers instead.

That does not seem right.

The code should only use the Received headers if List-Post is missing.

I'm not entirely sure there's even a use case for handling a missing List-Post header.


sebbASF:
Ping

sebbASF:
Code should not use Received headers. New PR to follow.

PR #402, opened 2017-09-06
Adding rundir to the devops config

cthistle:
Adding rundir to the devops config


gmcdonald:
I don't think we still need this.

PR #411, opened 2017-09-10
Drop unused items; don't mix tabs and spaces

sebbASF:


sebbASF:
Dropped the change to the pidfile arg because it is a separate issue

PR #416, opened 2017-09-12
INFRA-12732 - Add a label on PR

Humbedooh:
INFRA-12732 - Add a label on PR


Humbedooh:
seems like permissions are needed...

PR #428, opened 2017-09-14
Oops: need to quote back reference introducer

sebbASF:
Unfortunately this bug was not spotted earlier.


sebbASF:
Very strange: the diff shows that the original change did use a second backslash:

https://github.com/sebbASF/infrastructure-puppet/commit/7c49ec23a1976449f18759a9181bf584d5d8abc4#diff-938c82e29af80dc3f834d0985642368c

See lines 83-84

PR #439, opened 2017-09-23
Only parse headers

sebbASF:
No point parsing the entire message since only some of the headers are used


clambertus:
I'd like @Humbedooh to review this, I'm not well versed in the code.

sebbASF:
Ping

sebbASF:
PING

PR #472, opened 2017-10-04
INFRA-15193 - mboxer/archive.py does not work for mail aliases

sebbASF:
Added --lid option to allow list mail address to be specified
e.g. for aliases which don't have a List-Post or where the List-Post is incorrect


sebbASF:
Ping

PR #497, opened 2017-10-13
add celix-site to gitpubsub for when the ticket comes in :)

pono:


gmcdonald:
Please ping @pono when the ticket comes in then ...

gmcdonald:
should we be worried or chasing someone up that we are still waiting after 27 days?

gmcdonald:
Now 46+ days without any sign of a ticket coming in. Lets not preempt projects requests. They can create a new PR or Jira ticket when they are ready.

PR #521, opened 2017-10-22
default to having JIRA mirroring enabled

Humbedooh:
default to having JIRA mirroring enabled

### DIFF
--- a/modules/gitbox/files/cgi-bin/issues.cgi
+++ b/modules/gitbox/files/cgi-bin/issues.cgi
@@ -314,7 +314,7 @@ def main():
     project = "infra" # Default to infra
     if m:
         project = m.group(1)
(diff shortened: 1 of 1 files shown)


modules/gitbox/files/cgi-bin/issues.cgi
```diff
@@ -360,7 +360,7 @@ def main():
         sendEmail(mailto, email['subject'], email['message'])
 
     # Now do JIRA if need be
-    jiraopt = gconf.get('apache', 'jira') if gconf.has_section('apache') and gconf.has_option('apache', 'jira') else None
+    jiraopt = gconf.get('apache', 'jira') if gconf.has_section('apache') and gconf.has_option('apache', 'jira') else 'default'
```
gstein: drop the .has_section() as that is tested within the .has_option()

Consider using the Defaults mechanism for gconf/ConfigParser, so you don't even have to test if the option has been defined. It would add clarity to the script if you make the defaults explicit using that mechanism.

PR #536, opened 2017-10-26
Add entru for SDAP Incubating

lewismc:


clambertus:
looks like you set up a git repo for the website, what is the purpose of this commit?

lewismc:
@clambertus so we can host the Website code at Apache as well. Should this go somewhere else? we are looking at implementing GitPubSub. Thanks.

clambertus:
gitpubsub and git repos use a different authentication system (LDAP). this commit only updates auth for subversion repos, which should not be necessary for a git-based website. you will need to submit an infra jira to get gitpubsub set up for your website repo.

lewismc:
OK doke @clambertus so I this ticket should be closed off?

gmcdonald:
Yeah this PR is not required.

PR #549, opened 2017-11-07
The non-test list aliases are all private

sebbASF:
Treasurer, president, ea and secretary are all private


sebbASF:
Ping

Once puppet has run, someone will have to run newaliases on the host please (see INFRA-15325 for background)

sebbASF:
Ping?

PR #637, opened 2017-12-22
Add option for mails to be accessed by all committers

sebbASF:
This is intended for archiving mails to which all committers should have access.

e.g. the committers@apache.org archive

### DIFF
--- a/modules/mboxer/files/tools/archive.py
+++ b/modules/mboxer/files/tools/archive.py
@@ -27,6 +27,7 @@ Arguments (optional):
     <security-realm> (optional)
     - restricted - file the mail under the directory defined by the 'restricteddir' config item
     - private    - file the mail under the directory defined by the 'privatedir' config item
+    - committers - file the mail under the directory defined by the 'committersdir' config item
     - anything else, file it under the directory defined by the 'archivedir' config item
 
    The above can be combined if required.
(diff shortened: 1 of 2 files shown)


sebbASF:
Ping?

Humbedooh:
This poses a question of who we want accessing the box. Perhaps a topic for the ML. I'd be inclined to think we should only allow ASF members on there, on a per-request basis.

Humbedooh:
Come to think of it...what is the point of this? non-committers won't have access to the box in any case.

sebbASF:
The point is to potentially allow committers to have access to the committers@apache.org mail archives and any other lists that are open to all committers.

Humbedooh:
Committers won't have access to mbox-vm.

sebbASF:
Seems to me that the requirements for the archiver box need to be agreed on first.

gmcdonald:
I'm also going to say Im a -1 to this change. mbox-vm itself is not meant to be accessed.

gstein:
mbox-vm was constructed to provide ASF Members with access to our archives, based on the general precedent of Members having access to the records of the Foundation (caveat of PII types of email).

Committers are subscribed to committers@ and can retain their email. The Foundation will not provide access to mbox-vm to review/recover those emails. In general, access to mbox-vm will be "upon request" rather than "automatic" for Members. It will be the "source of truth" for our email archives, and will (thus) be reasonably tight on permissions/access.

PR #691, opened 2018-01-18
also install liblzma-dev on mergebot-vm2

jasonkuster:
@clambertus take two


jasonkuster:
hm, never mind -- it's already installed? I'll dig into this further.

PR #701, opened 2018-01-19
Pointing to new server.

cthistle:
Pointing to new server.


cthistle:
Don't merge till 0400 UTC as per @Humbedooh

PR #703, opened 2018-01-20
Accept $baseurl/repo:branch (thus allowing non-standard branches)

Humbedooh:
Accept $baseurl/repo:branch (thus allowing non-standard branches)


clambertus:
travis fail seems to be due to a bug in lint.sh

PR #809, opened 2018-03-15
add apache mod headers and disable a2enmod for ofbiz

pono:


JacquesLeRoux:
Thank you for that guys, I did not read https://enable-cors.org/server_apache.html right and used only a2enmod headers. 
CORS with Auth header is really stressful from JavaScript side I found

PR #815, opened 2018-03-19
Allow project owners to read their incubator archives

sebbASF:
Also speeds up the script by only reading LDAP once


sebbASF:
Could this be merged or rejected please?

PR #820, opened 2018-03-20
Allow ASF members to view /pmc/ URLs

sebbASF:
As per INFRA-15650, use RequireAny to group PMC and ASF members

Note that there is no need to change the definition in the gen-httpdconfig.sh file, as that is no longer used (it was intended for dist.a.o)

### DIFF
--- a/data/nodes/mailprivate-vm.apache.org.yaml
+++ b/data/nodes/mailprivate-vm.apache.org.yaml
@@ -90,20 +90,16 @@ vhosts_asf::vhosts::vhosts:
       <LocationMatch ^/pmc/private-arch/$dir(.*)>
         ## Either PMC member or ASF member
         <RequireAny>
+            AuthType Basic
+            AuthBasicProvider ldap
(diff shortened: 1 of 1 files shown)


clambertus:
Preliminary testing on this doesn't work. I think it's related to the difference in ldapUrl necessary for a query against memberUid vs a query against owner=. I have the code in place on mailprivate-vm currently if you want to take a look.

sebbASF:
Strange. I just added the check from the MacroMboxMembers macro, so I would expect it to work.
Maybe there is an httpd bug?

sebbASF:
I think MacroMboxMembers must be wrong.
It is supposed to be used for directories that don't match projects. However it doesn't allow access and it does not work for mail-search either.

clambertus:
I've temporarily applied this on the server for review. Could you please also correct the missing closing / on line 111.

sebbASF:
Works for me.

clambertus:
This has now been applied on the server. Great job in getting this working, thanks!

PR #827, opened 2018-03-22
Try again to fix INFRA-15650

sebbASF:
This needs more work to enable the AuthzProviderAlias section.
However it should be usable as is (except ASF members won't get access to the /pmc area).
But at least PMC members will have access to the /pmc area


gmcdonald:
This is fine by me, what we end up with currently is still an marked improvement on mail-search as it was.

PR #842, opened 2018-03-27
Propose quick fix pylint/typos

jeis2497052:
are these updates OK?

### DIFF
--- a/modules/gitbox/files/matt/tools/grouper.py
+++ b/modules/gitbox/files/matt/tools/grouper.py
@@ -15,19 +15,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-import sys
-import re
-import ldap
-import urllib2
-import json
-import requests
-import hashlib
 import ConfigParser
-import sqlite3
+import json
 import logging
+import os
+import re
+import sqlite3
+import sys
+import urllib2
 
-logging.basicConfig(filename='grouper.log', format='[%(asctime)s]: %(message)s', level=logging.INFO)
+import ldap
+import requests
+
+logging.basicConfig(filename='grouper.log',
+                    format='[%(asctime)s]: %(message)s', level=logging.INFO)
 
 # LDAP Defs
 UID_RE = re.compile("uid=([^,]+),ou=people,dc=apache,dc=org")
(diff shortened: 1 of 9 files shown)


gstein:
ezt.py is upstream. No patches allowed on that (create a PR at https://github.com/gstein/ezt) ... @Humbedooh is working on a patch to place ezt on all boxes, so that we can simply remove local copies from our modules.

The other changes generally look good. It would be great if you could back out the ezt change, and then we roll forward with the rest of your changs. Thanks!

jeis2497052:
sorry about ezt, I will back that out shortly

modules/gitbox/files/matt/tools/grouper.py
```diff
@@ -15,29 +15,30 @@
...
+                    format='[%(asctime)s]: %(message)s', level=logging.INFO)
 
 # LDAP Defs
 UID_RE = re.compile("uid=([^,]+),ou=people,dc=apache,dc=org")
 
 # Run `python grouper.py debug` to check teams but not add/remove users
 DEBUG_RUN = True if len(sys.argv) > 1 and sys.argv[1] == 'debug' else False
 if DEBUG_RUN:
-    print("Debug run active! Not modifying teams")
+    print "Debug run active! Not modifying teams"
```
gstein: Oh, I think that grouper.py might run under Py3, so this tweak would be a problem.

gstein:
Thanks, @jeis2497052 ! ... very happy to see new faces contributing! Looks like a small tweak, and this can be merged.

Much appreciated!

gstein:
@gmcdonald hashlib isn't used in the file, so removing that import is Good(tm) :-)

jeis2497052:
Great!  so just add the parens for Py2 back then ?  sorry about the parens, I keep forgetting they are required in some versions Py2.

gstein:
@jeis2497052 the parens are needed for Py3, as "print" switched from a statement to a function. For Py2, the parens are essentially a no-op.

Then we merge! :-)

gstein:
Fabulous. Thanks, John! ... More, please :-)

jeis2497052:
Happy to contribute.  Thanks for walking me through the upstream issues and paren issues.

PR #855, opened 2018-04-04
Add trial redirect for argus => ranger

sebbASF:
Start with a single redirect

Also added some comments for later maintainers


gmcdonald:
If this is going to end up being a big list, I wonder if a macro would help, and maintain a separate file with the from -> to list?

sebbASF:
It might be worth using a separate conf file for the common settings.
It could then be included by both the http and https sections.
I don't think a macro would be better here.

sebbASF:
The argus redirect works.

PR #863, opened 2018-04-08
INFRA-15966 - renamed projects and lists

sebbASF:
Add renames for projects and lists


gmcdonald:
Exactly what I asked for , thanks!

sebbASF:
(See also #855)
Could not work out how to use puppet to create an include file to be included in the vhosts.
custom_config is not suitable as the created file is automatically included before the vhosts.
Whereas Macros can be defined before vhosts and 'included' as necessary.

PR #869, opened 2018-04-11
Comments not allowed on macro invocations

sebbASF:


sebbASF:
Needs rework as part has been applied

gmcdonald:
yeah sorry about that, mail-archives was down, and I did not see this at the time.

PR #870, opened 2018-04-11
Infra 15754 - fix year end processing

sebbASF:
Also fixes misplaced exit statement


sebbASF:
Note: the file now agrees with

https://svn.apache.org/repos/asf/httpd/mod_mbox/trunk/scripts/update-index-monthly

However gzcat has been changed to zcat.

gmcdonald:
That's fine I guess. I've had other scripts determine which to use depending on OS, but not really required here.

gmcdonald:
Are you happy for this to be merged now?

sebbASF:
Yes, fine to be merged.
My comment was just intended to note that the local copy now agrees with the original apart from the necessary zcat change. (I've raised a bug for the hard-coded name).

PR #878, opened 2018-04-12
Synchronise layout

sebbASF:
themis and tlp-eu-west have almost the same content.
To make it easier to compare them, it helps if the spacing and
ordering is the same.

No functional changes.


sebbASF:
ping?

PR #883, opened 2018-04-14
INFRA-16088

clambertus:
switch wcsub location for maven and maven/doxia


clambertus:
self-merging due to ongoing coordination with Herve.

PR #884, opened 2018-04-14
INFRA-16088

clambertus:


gmcdonald:
Happy to merge this when we are both around

PR #901, opened 2018-04-25
alter cron jobs for authz generation

gstein:
* add cron job for the new python-based authz generation
* remove the "ldap_change" authz step, as it is redundant with the
  other generation step

note: the old perl-based "template_commit" generate is still running
and creates proper files (the new py step creates *.alt for now)

### DIFF
--- a/modules/subversion_server/manifests/init.pp
+++ b/modules/subversion_server/manifests/init.pp
@@ -649,15 +649,16 @@ class subversion_server (
       hour    => '*',
       user    => 'www-data',
       command => '/x1/svn/scripts/authorization/gen_asf-authorization.pl template_commit > /dev/null';
+    'authz-rebuild-new':
+      minute  => '*/3',
+      hour    => '*',
+      user    => 'www-data',
+      command => '/x1/svn/scripts/authorization/gen.py';
     'generate-dist-auth':
       minute  => '15',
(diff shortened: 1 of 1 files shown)


modules/subversion_server/manifests/init.pp
```diff
@@ -649,15 +649,16 @@
       hour    => '*',
       user    => 'www-data',
       command => '/x1/svn/scripts/authorization/gen_asf-authorization.pl template_commit > /dev/null';
+    'authz-rebuild-new':
+      minute  => '*/3',
+      hour    => '*',
+      user    => 'www-data',
+      command => '/x1/svn/scripts/authorization/gen.py';
```
gmcdonald: Where is this file in puppet ?

gstein: Euh... https://github.com/apache/infrastructure-puppet/blob/deployment/modules/subversion_server/files/scripts/authorization/gen.py

Sibling to /x1/svn/scripts/authorization/gen_asf-authorization.pl which is also in that scripts/authorization directory.

gmcdonald: gdi, I was in the wrong auth dir

modules/subversion_server/files/authorization/
 
nm then

PR #909, opened 2018-04-28
INFRA-16402 myfaces-master-pom moved to gitbox

deki:


deki:
@cthistle could you please merge it to avoid further commits in SVN?

PR #914, opened 2018-04-30
Add Perl JSON package to base ubuntu.

sebbASF:
Also trim trailing spaces and fix alpha order issue


pono:
i < l, so I think firefox is in the right place.
Otherwise looks good!

pono:
Thanks Sebb!

PR #919, opened 2018-05-01
let svn manage the dist/zzz/ directory

gstein:
if puppet tries to manage it, then svn will get all complain-y


gstein:
not needed. the puppet line only modifies when the dir is present, and svn won't notice/care about the chown.

PR #936, opened 2018-05-08
Added ant 1.10.3 and rebuild the way the 'latest' sysmlinks are created and updated.

cthistle:
Added ant 1.10.3 and rebuild the way the 'latest' sysmlinks are created and updated.


cthistle:
Started off just adding ant, turned out updating an existing symlink doesn't work correctly so I rebuilt it.

PR #982, opened 2018-05-19
Standardise python hash-bang

sebbASF:


sebbASF:
Ping: this standardises the Python hash-bangs

PR #1008, opened 2018-06-08
Treat www-eu and www-us equally

sebbASF:
AFAICT it does not make sense to only support www-us for the (shared) tlpserver.

Applying this should fix the broken www-eu.a.o URLs.

### DIFF
--- a/data/roles/tlpserver.yaml
+++ b/data/roles/tlpserver.yaml
@@ -415,6 +415,7 @@ vhosts_asf::vhosts::vhosts:
       - 'apachegroup.org'
       - 'www.apachegroup.org'
       - 'www.*.apache.org'
+      - 'www-eu.apache.org'
       - 'www-us.apache.org'
     directories:
       -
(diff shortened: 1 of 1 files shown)


henkp-apache:
AFAICT it does not make sense to support www-{eu,us} for the (shared) tlpservers ;
www-{eu,us}.a.o URLs only appeared on generated pages ; those pages are now gone.

sebbASF:
I see two problems with dropping www-eu.a.o and www.us.a.o:

1) www.[eu|us].a.o do not support SSL

2) Having introduced www-[eu|us] to deal with the SSL issue, it's not right to arbitrarily stop supporting the hostname.

Humbedooh:
I'm okay with supporting it in the short run.
In the long run, we'll likely figure out a way of not using either of these solutions, and go with a download.a.o instead.

sebbASF:
Thanks. That has fixed the www-eu.a.o URL.

Note that it's not just about supporting downloads; the URLs can also be used for accessing the ASF website.

PR #1010, opened 2018-06-10
Copy new ASF favicon from modules/mail_archives/files/favicon.ico

sebbASF:
Replace old ASF favicons


sebbASF:
Can this be applied please?

PR #1019, opened 2018-06-12
INFRA-16511 Use unqualified host for Attic header

sebbASF:


Humbedooh:
/var/www is a symlink to /www, so shouldn't need a change

sebbASF:
Oops - did not intend to change the path spec; reverted that part

Humbedooh:
I'm just testing a thing on a PR, ignore

sebbASF:
PING

PR #1022, opened 2018-06-13
Adding logrotate config for Apache2 on netbeans-vm.apache.org

jpirek:
Hello,
for netbeans-vm.apache.org we need Apache2 logfiles to rotate daily so we can do some ad-hoc processing and analysis of daily logs.


gmcdonald:
Thanks @jpirek !

jkovalsky:
Thanks @gmcdonald !

PR #1085, opened 2018-07-12
hpp-attic-171 ; retire lucy

henkp-apache:


gmcdonald:
Im happy with the changes Henk, however, I prefer you remove the entries instead of commenting them out, lets do that and then I'll merge.

PR #1090, opened 2018-07-14
change cfgscanner to systemd service, run in foreground

Humbedooh:
change cfgscanner to systemd service, run in foreground


Humbedooh:
wrong mod!

PR #1178, opened 2018-09-27
INFRA-17067

clambertus:
Fix content-type for signature files

### DIFF
--- a/data/nodes/memnon.apache.org.yaml
+++ b/data/nodes/memnon.apache.org.yaml
@@ -67,6 +67,7 @@ vhosts_asf::vhosts::vhosts:
       LuaScope thread
       LuaCodeCache stat
       IndexOptions FoldersFirst FancyIndexing
+      AddType text/plain .asc .md5 .sha .sha1 .sha2 .sha256 .sha512 .sha3 .mds
       
   archive_https:
     vhost_name: '*'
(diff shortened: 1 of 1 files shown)


data/nodes/memnon.apache.org.yaml
```diff
@@ -67,6 +67,7 @@ vhosts_asf::vhosts::vhosts:
       LuaScope thread
       LuaCodeCache stat
       IndexOptions FoldersFirst FancyIndexing
+      AddType text/plain .asc .md5 .sha .sha1 .sha2 .sha256 .sha512 .sha3 .mds
```
sebbASF: BTW, the type is correct for .asc files currently, so I assume there must be another definition somewhere?
Should they be merged?

clambertus: same block we use on svn/dist copypasta

sebbASF: Yes, I know.
But prior to the addition, .asc was working OK, so there must be some other place where that is defined. Seems to me it would be best to have all the types defined in the same place.

sebbASF: No, I meant it was *already* working previously - and it still is.

clambertus: Got it, thanks. The asc content type is a default in httpd. The fix I committed is the 'correct' method for updating content type. If httpd internals already set it to text/plain, then my change is a no-op, and does actually "have all the types defined in the same place."

PR #1199, opened 2018-10-06
hpp-dist-htaccess

henkp-apache:
Add dist/.htaccess ; += .tgz


henkp-apache:
re: https://issues.apache.org/jira/browse/INFRA-15707

PR #1217, opened 2018-10-24
adding blocky to bz and archive.a.o

cthistle:
adding blocky to bz and archive.a.o


data/nodes/asf-bz1-us-mid.priv.apache.org.yaml
```diff
@@ -6,6 +6,7 @@ classes:
   - apache::mod::remoteip
   - bugzilla
   - bugzilla::custom_project
+  - blocky
```
Humbedooh: Not needed here (it's enabled on the proxy in front), and possibly adding stuff to an already burdened box.

PR #1226, opened 2018-10-30
Change camel-website checkout directory path

zregvart:
Makes the checkout directory for the camel-website.git:asf-site branch
be `/www/camel.apache.org/content/staging` as the current site has a
`/content` which is matched by rewrite rules. As suggested by Daniel
Gruno[1]

[1] https://issues.apache.org/jira/browse/INFRA-15969?focusedCommentId=16668569#comment-16668569


zregvart:
Thanks!

PR #1230, opened 2018-11-06
adding jenkins02 specific snmpd config

cthistle:
adding jenkins02 specific snmpd config


gmcdonald:
Can I get some history about this please, why it is needed etc ...

Humbedooh:
it's a specific jenkins software raid check we have.

cthistle:
while cleaning up Puppet errors, I found /etc/snmp/snmpd.conf appended with "dlmod swRaidMIB /usr/lib/swRaidPlugin.so" and marked immutable. This is just me trying to make that go.

PR #1250, opened 2018-11-25
[INFRA-17298] mod_proxy_wstunnel is added

solomax:


martin-g:
Don't we also need:

https://github.com/apache/infrastructure-puppet/blob/13524566d230dd9db18fdeb244fb6703d9dc214d/data/nodes/openmeetings-vm2.apache.org.yaml#L39-L41

https://github.com/apache/infrastructure-puppet/blob/13524566d230dd9db18fdeb244fb6703d9dc214d/data/nodes/openmeetings-vm2.apache.org.yaml#L73-L74

?

martin-g:
The PR description should probably mention `INFRA-17298`

solomax:
@martin-g I'm not sure :( is this is required or not
Our config uses http <-> https and ws <-> wss
Will try to perform some tests on our internal server first, if I do remember correctly simple ws <-> ws proxy might work as is ...

PR #1270, opened 2018-12-21
Add a simple JIRA-To-Slack bridge on the jira box

Humbedooh:
Add a simple JIRA-To-Slack bridge on the jira box


modules/jira_slack_bridge/templates/jira-slack-bridge.py.erb
```diff
@@ -0,0 +1,132 @@
...
+        sendNotice(js)
+
+def main():
+    # Keep track of the last change and ticket we saw, so we know what's new
+    # This is determined on first run.
+    lastChangeID = 0
+    lastJIRA = "INFRA-0"
+    firstRun = True
+    while True:
+        js = requests.get(CHGURL, auth = ('<%= @jirauser %>', '<%= @jirapass %>')).json()
```
gstein: For clarity, I think you should move ALL substitutions to the top, as global constants. In normal Python code, they'd be pejoratively called "magic constants". I think they're even worse as "magic substituted contents".

Humbedooh: Done!

PR #1426, opened 2019-01-30
infrastructure-cvs@apache.org => commits@infra.apache.org

sebbASF:
Replace old list name (alias) with actual list name
This might help stop duplicate emails for commits to /META arriving at commits@


sirisak1985:
gem install bundler
bundle install

PR #1430, opened 2019-02-05
adding all commons repos to auth.cfg INFRA-17724

cthistle:
adding all commons repos to auth.cfg INFRA-17724


cthistle:
merging my own PR as per +1 from humbedooh in Slack

PR #1439, opened 2019-02-21
There are two more sitemap index files

sebbASF:
robots.txt is supposed to list all the sitemap index files; there are now two more


sebbASF:
Ping?

PR #1441, opened 2019-02-25
removing searchdictionary as per INFRA-17850

cthistle:
removing searchdictionary as per INFRA-17850


snoopdave:
thanks @cthistle and @Humbedooh

PR #1443, opened 2019-02-26
clarify change_type

gstein:
Clarify how change_type is derived, and allow for an unknown state (instead of throwing a KeyError)


cybernet:
no one wants to review this ?

PR #1446, opened 2019-02-28
testing multimail branching

Humbedooh:


Humbedooh:
bla bla testing with new subject line :)

gmcdonald:
not doing this now

PR #1455, opened 2019-03-05
Fix bug: missing op variable; also fix path concat bug

sebbASF:
See INFRA-17958


gmcdonald:
The common.conf file does exist on hermes-vm - this module is designed  for hermes-vm, not the current hermes which I believe you are trying to fix. 

Please just commit using svn your required changes for hermes.

PR #1461, opened 2019-03-07
checking if IP exists before doing more stuff

cthistle:
checking if IP exists before doing more stuff


cthistle:
+1 from humbedooh via slack

PR #1465, opened 2019-03-08
Add IPv6 to blocky2, tweak some things as per greg

Humbedooh:
Add IPv6 to blocky2, tweak some things as per greg

### DIFF
--- a/modules/blocky/files/blocky.py
+++ b/modules/blocky/files/blocky.py
@@ -36,6 +36,8 @@ SYSLOG = None
 def getbans(chain = 'INPUT'):
    """ Gets a list of all bans in a chain """
    banlist = []
+   
+   # Get IPv4 list
    tries = 0
    while tries < 10:
       tries += 1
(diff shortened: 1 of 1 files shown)


modules/blocky/files/blocky.py
```diff
@@ -65,19 +67,61 @@ def getbans(chain = 'INPUT'):
...
+                  'protocol': protocol,
+                  'option': '---',
+                  'source': source,
+                  'destination': destination,
+                  'extensions': extensions
+               }
+               
+               banlist.append(entry)
+         break
       except subprocess.CalledProcessError as err:
```
gstein: please put the `except:` around the `check_output()` call, moving the processing of `out` outside of the try/except block

modules/blocky/files/blocky.py
```diff
@@ -65,19 +67,61 @@ def getbans(chain = 'INPUT'):
...
+         if 'you must be root' in err.output:
+            print("Looks like blocky doesn't have permission to access iptables, giving up completely! (are you running as root?)")
+            sys.exit(-1)
+         time.sleep(1) # write lock, probably
+   # Get IPv6 list
+   tries = 0
+   if not os.path.exists("/sbin/ip6tables"):
+      return banlist
+   while tries < 10:
+      tries += 1
```
gstein: if `tries` is unused, then you could do `for _ in range(10):` since the code does a `break` upon success. you can also add an `else` to the for-loop to syslog that the retry count was exceeded.

gstein: I might suggest a global such as `RETRY_COUNT`

modules/blocky/files/blocky.py
```diff
@@ -65,19 +67,61 @@ def getbans(chain = 'INPUT'):
...
          time.sleep(1) # write lock, probably
+   return banlist
       
 def iptables(ip, action):
     """ Runs an iptables action on an IP (-A, -C or -D), returns true if
         succeeded, false otherwise """
     try:
+        exe = "/sbin/iptables"
+        if ':' in ip:
+            exe = "/sbin/ip6tables"
```
gstein: maybe some global constants for these paths? not really needed for correctness, but maybe clarity.

PR #1472, opened 2019-03-13
renaming the current hudi repo as it's actually the site

cthistle:
renaming the current hudi repo as it's actually the site


cthistle:
nevermind, just going to merge asf-site branch into their existing incubator-hudi repo.

PR #1481, opened 2019-03-20
INFRA-18055

dfoulks1:
Added jira notifications for the NEMO project to ASF slack channel #nemo-notifications


modules/jira_slack_bridge/files/slackbridge.yaml
```diff
@@ -5,3 +5,6 @@ subscriptions:
     guacamole:
         channel: "#guacamole"
         jira: GUACAMOLE
+    nemo:
+	channel: "#nemo-notifications"
```
wonook: It seems different from normal whitespaces. Maybe this is preventing it from working?

PR #1488, opened 2019-03-23
INFRA-18067  comment out fields that are not used

clambertus:


clambertus:
the ICLA fields are "used" just not editable by the user

PR #1491, opened 2019-03-24
Revert "Added link to view commit diff online"

clambertus:
Reverts apache/infrastructure-puppet#1489


clambertus:
Had to revert because this commit somehow broke gitmail entirely

PR #1508, opened 2019-03-30
More System/Help pages to exclude

sebbASF:


gmcdonald:
should we wait here until some files are added for exclusion ?

sebbASF:
Oops, forgot to write the data out...

PR #1527, opened 2019-04-17
Add Maven 3.6.1

rfscholte:
See https://maven.apache.org/docs/3.6.1/release-notes.html
Especially https://issues.apache.org/jira/browse/WAGON-537 is interesting, it'll speed up downloads and uploads a lot


cybernet:
no updates?

cthistle:
I'll get that up in Bintray, then there are a few more places this needs updated in the config.

PR #1529, opened 2019-04-21
Remove .incubator from directory name.

sebbASF:
*** 
Ideally this patch should be committed over a month end to reduce the number of mailboxes that will have to be merged. 
i.e. at around midnight UTC on April 30th 
***

See INFRA-18153.

This patch removes the .incubator. segment from the directory name on mod_mbox so mail-box directory names don't change when podlings graduate.

### DIFF
--- a/modules/mboxer/files/tools/archive.py
+++ b/modules/mboxer/files/tools/archive.py
@@ -18,8 +18,10 @@
 """ 
 
 This script determines the intended ASF recipient of an email and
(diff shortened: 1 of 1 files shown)


sebbASF:
In retrospect, I think it would be better to fix this now so it has time to bed in.

Whilst waiting until the end of the month may avoid some mailboxes being split across the different directories, it won't help with all the existing splits due to graduation, and merging of the boxes might not even be necessary depending on the chosen minotaur cutover month.

PR #1549, opened 2019-05-09
bumping gradle version to 4.10.3 as 4.10.2 seems to have gone missing

cthistle:
bumping gradle version to 4.10.3 as 4.10.2 seems to have gone missing


cthistle:
Will need to update the cwiki page and builds.a.o once this is merged.

PR #1559, opened 2019-05-21
adding some error checking to the backup scripts

cthistle:
adding some error checking to the backup scripts

### DIFF
--- a/modules/mysql_asf/templates/dbsave_mysql.sh.erb
+++ b/modules/mysql_asf/templates/dbsave_mysql.sh.erb
@@ -29,7 +29,7 @@ RSYNC_OFFSITE="<%= @rsync_offsite %>"
 backup_db() {
   echo "Dumping ${DBNAME} DB..."
   time ${DUMP} -h ${DBHOST} -u ${DBUSER} --password=${PASS} -P ${PORT} ${DBNAME} | gzip --rsyncable -9 > ${DUMPFILE}
-  if [ ${PIPESTATUS[0]} -ne 0 ]
+  if [[ ${PIPESTATUS[@]} != "0 0" ]]
     then
       ERROR_FOUND=1
   fi
--- a/modules/mysql_asf/templates/dbsave_mysql.sh.erb
+++ b/modules/mysql_asf/templates/dbsave_mysql.sh.erb
@@ -29,7 +29,7 @@ RSYNC_OFFSITE="<%= @rsync_offsite %>"
 backup_db() {
   echo "Dumping ${DBNAME} DB..."
   time ${DUMP} -h ${DBHOST} -u ${DBUSER} --password=${PASS} -P ${PORT} ${DBNAME} | gzip --rsyncable -9 > ${DUMPFILE}
-  if [ $? -eq 1 ]
+  if [ $? -ne 0 ]
     then
       ERROR_FOUND=1
   fi
@@ -54,7 +54,7 @@ create_dumpdir() {
   if [ ! -d ${DUMPDIR} ]; then
     echo "Creating new directory structure ( ${DUMPDIR} ) for new DB ${DBNAME}"
     mkdir -p ${DUMPDIR};
-    if [ $? -eq 1 ]
+    if [ $? -ne 0 ]
       then
         ERROR_FOUND=1
     fi
@@ -82,7 +82,7 @@ rsync_offsite() {
   --log-file=/root/rsynclogs/abi/backups-<%= @hostname %>-$TODAY.log \
   --password-file=/root/.pw-abi \
   ${DUMPFILE} rsync://<%= @rsync_user %>@localhost:1873/<%= @hostname %>/
-  if [ $? -eq 1 ]
+  if [ $? -ne 0 ]
   then
     ERROR_FOUND=1
   fi
--- a/modules/mysql_asf/templates/dbsave_mysql.sh.erb
+++ b/modules/mysql_asf/templates/dbsave_mysql.sh.erb
@@ -132,9 +132,4 @@ done
 stop_lock
 date
 
-if [ ERROR_FOUND -eq 1 ]
-  then
-    exit 1
-  else
-    exit 0
-fi
+exit $ERROR_FOUND
--- a/modules/mysql_asf/templates/dbsave_mysql.sh.erb
+++ b/modules/mysql_asf/templates/dbsave_mysql.sh.erb
@@ -29,7 +29,7 @@ RSYNC_OFFSITE="<%= @rsync_offsite %>"
 backup_db() {
   echo "Dumping ${DBNAME} DB..."
   time ${DUMP} -h ${DBHOST} -u ${DBUSER} --password=${PASS} -P ${PORT} ${DBNAME} | gzip --rsyncable -9 > ${DUMPFILE}
-  if [ $? -ne 0 ]
+  if [ ${PIPESTATUS[0]} -ne 0 ]
     then
       ERROR_FOUND=1
   fi
--- a/modules/mysql_asf/templates/dbsave_mysql.sh.erb
+++ b/modules/mysql_asf/templates/dbsave_mysql.sh.erb
@@ -13,6 +13,7 @@ OLDDUMPDATE=`${DATE} +%Y%m%d%H%M --date='last month'`
 DUMPOPTS='--lock-tables --master-data=1'
 DUMPROOT='/x1/db_dump/mysql'
 DUMPROOT=<%= @dumproot %>
+EXIT_STATUS=0
 LOCKFILE='/tmp/mysql.dump.lock'
 MYSQL=$(which mysql)
 PASS=$(cd /root/ ; cat .my.cnf | grep -i -m 1 pa | sed 's/password=//g' | sed "s/'//g")
@@ -31,7 +32,7 @@ backup_db() {
   time ${DUMP} -h ${DBHOST} -u ${DBUSER} --password=${PASS} -P ${PORT} ${DBNAME} | gzip --rsyncable -9 > ${DUMPFILE}
   if [[ ${PIPESTATUS[@]} != "0 0" ]]
     then
-      ERROR_FOUND=1
+      EXIT_STATUS=1
   fi
   echo -e "\n\n"
 }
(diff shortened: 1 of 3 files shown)


modules/mysql_asf/templates/dbsave_mysql.sh.erb
```diff
@@ -29,6 +29,10 @@ RSYNC_OFFSITE="<%= @rsync_offsite %>"
 backup_db() {
   echo "Dumping ${DBNAME} DB..."
   time ${DUMP} -h ${DBHOST} -u ${DBUSER} --password=${PASS} -P ${PORT} ${DBNAME} | gzip --rsyncable -9 > ${DUMPFILE}
+  if [ $? -eq 1 ]
```
sebbASF: AFAICT this will only pick up errors in gzip, not in the dump command

Furthermore, most commands return 0 on success and >0 otherwise, so comparison with 1 is not guaranteed to catch all errors

modules/mysql_asf/templates/dbsave_mysql.sh.erb
```diff
@@ -50,6 +54,10 @@ create_dumpdir() {
   if [ ! -d ${DUMPDIR} ]; then
     echo "Creating new directory structure ( ${DUMPDIR} ) for new DB ${DBNAME}"
     mkdir -p ${DUMPDIR};
+    if [ $? -eq 1 ]
```
sebbASF: Should check for non-zero (or > 0) here

modules/mysql_asf/templates/dbsave_mysql.sh.erb
```diff
@@ -74,6 +82,11 @@ rsync_offsite() {
   --log-file=/root/rsynclogs/abi/backups-<%= @hostname %>-$TODAY.log \
   --password-file=/root/.pw-abi \
   ${DUMPFILE} rsync://<%= @rsync_user %>@localhost:1873/<%= @hostname %>/
+  if [ $? -eq 1 ]
```
sebbASF: Ditto

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
```diff
@@ -25,6 +26,10 @@ backup_db() {
   echo ""
   echo "Dumping ${DBNAME} DB..."
   time su - <%= @user %> -c "${DUMP} ${DUMPOPTS} ${DBNAME}" | gzip --rsyncable -9 > ${DUMPFILE}
+  if [ $? -eq 1 ]
```
sebbASF: As above, this does not check the status of the dump command.
Also, 1 is not the only possible error code

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
```diff
@@ -66,6 +71,11 @@ rsync_offsite() {
   --log-file=/root/rsynclogs/abi/backups-titan-postgres-$TODAY.log \
   --password-file=/root/.pw-abi \
   ${DUMPFILE} rsync://apb-titan@localhost:1873/titan/
+  if [ $? -eq 1 ]
+    then
```
sebbASF: As before

modules/rsync_asf/templates/rsync-offsite.sh.erb
```diff
@@ -22,6 +23,11 @@ rsync_offsite() {
   --log-file=/root/rsynclogs/abi/backups-<%= @hostname %>-$TODAY.log \
   --password-file=/root/.pw-abi \
   ${FSNAME} rsync://apb-<%= @hostname %>@localhost:1873/<%= @hostname %>/
+  if [ $? -eq 1 ]
```
sebbASF: etc

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
cthistle: that's all one command, not piped, so that should work as expected.

sebbASF: By as before, here I meant that comparison with 1 is unsafe

modules/mysql_asf/templates/dbsave_mysql.sh.erb
cthistle: Updated in commit https://github.com/apache/infrastructure-puppet/pull/1559/commits/bfe4cee2504df73b6578bd8b58b8c182a32abd12

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
cthistle: Updated in commit https://github.com/apache/infrastructure-puppet/pull/1559/commits/bfe4cee2504df73b6578bd8b58b8c182a32abd12

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
cthistle: Updated in commit https://github.com/apache/infrastructure-puppet/pull/1559/commits/bfe4cee2504df73b6578bd8b58b8c182a32abd12

modules/rsync_asf/templates/rsync-offsite.sh.erb
cthistle: Updated in commit https://github.com/apache/infrastructure-puppet/pull/1559/commits/bfe4cee2504df73b6578bd8b58b8c182a32abd12

modules/mysql_asf/templates/dbsave_mysql.sh.erb
cthistle: fixed

modules/mysql_asf/templates/dbsave_mysql.sh.erb
```diff
@@ -119,3 +132,9 @@ done
 stop_lock
 date
 
+if [ ERROR_FOUND -eq 1 ]
```
sebbASF: Surely that should be "if [ $ERROR_FOUND -eq 1 ]" ?

But it would be simpler to just replace the conditional block with:

exit $ERROR_FOUND

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
```diff
@@ -106,3 +116,10 @@ done
 
 stop_lock
 date
+
+if [ ERROR_FOUND -eq 1 ]
```
sebbASF: As above

modules/rsync_asf/templates/rsync-offsite.sh.erb
```diff
@@ -32,3 +38,10 @@ for FSNAME in `cat "$DUMPLISTFILE"` ; do
   rsync_offsite
   sleep 5 
 done
+
+if [ ERROR_FOUND -eq 1 ]
```
sebbASF: As above

modules/mysql_asf/templates/dbsave_mysql.sh.erb
cthistle: makes sense, updated in 1a7c38602bee1aba23e091553918ba64503d921b

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
cthistle: updated in 1a7c38602bee1aba23e091553918ba64503d921b

modules/rsync_asf/templates/rsync-offsite.sh.erb
cthistle: updated in 1a7c38602bee1aba23e091553918ba64503d921b

cthistle:
@sebbASF , any more review/changes?

modules/mysql_asf/templates/dbsave_mysql.sh.erb
```diff
@@ -132,9 +132,4 @@ done
 stop_lock
 date
 
-if [ ERROR_FOUND -eq 1 ]
-  then
-    exit 1
-  else
-    exit 0
-fi
+exit $ERROR_FOUND
```
sebbASF: This change is fine, but I just noticed that ERROR_FOUND is not initialised.
Also, it might be better to rename it as EXIT_STATUS
And it could hold the actual status from the last failed command rather than 1

The script continues processing if  there is an error; I don't know if that is intended.
There does not seem to be any logging of the error, so determining what went wrong may be tricky

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
```diff
@@ -117,9 +117,4 @@ done
 stop_lock
 date
 
-if [ ERROR_FOUND -eq 1 ]
-  then
-    exit 1
-  else
-    exit 0
-fi
+exit $ERROR_FOUND
```
sebbASF: As above, except ERROR_FOUND is initialised OK

modules/rsync_asf/templates/rsync-offsite.sh.erb
```diff
@@ -39,9 +39,4 @@ for FSNAME in `cat "$DUMPLISTFILE"` ; do
   sleep 5 
 done
 
-if [ ERROR_FOUND -eq 1 ]
-  then
-    exit 1
-  else
-    exit 0
-fi
+exit $ERROR_FOUND
```
sebbASF: As above, except ERROR_FOUND is re-initialised for each rsync performed.
I.e. the script will only return the status of the last rsync.
Might be better to have two scripts - one to do an rsync of a single FSNAME (passed as a param) and another to loop round the list of names. This would be easier to test.

modules/mysql_asf/templates/dbsave_mysql.sh.erb
cthistle: Yes the script is intended to continue. Logging is going to Datadog, so we can alert when the script fails.

cthistle: I also agree that EXIT_STATUS makes more sense and I'll update that.

modules/postgresql_asf/templates/dbsave_postgres.sh.erb
cthistle: set EXIT_STATUS=0 at the start of the script

modules/rsync_asf/templates/rsync-offsite.sh.erb
cthistle: Whoops, put my declaration in the wrong spot. fixed

PR #1564, opened 2019-05-24
changing autocrlf to input INFRA-18383

cthistle:
changing autocrlf to input INFRA-18383

### DIFF
--- a/environments/windows/modules/jenkins_slave_windows/templates/gitconfig.txt.erb
+++ b/environments/windows/modules/jenkins_slave_windows/templates/gitconfig.txt.erb
@@ -1,6 +1,6 @@
 [core]
   symlinks = false
-  autocrlf = true
+  autocrlf = input
   fscache = true
   longpaths = true
 [color]


vlsi:
Unfortunately this change breaks Windows checkouts.
Typical Git configuration for WIndows is autocrlf=true (== Windows editors are configured to consume and produce CRLF line endings), and `autocrlf=input` produces unexpected results.

Tibor17:
@gmcdonald 
`core.autocrlf=input` is still wrong because it modifies the sources to use LF.
Git installation should not modify sources.
Therefore `core.autocrlf=false` should be used. Can you fix it?
In the Jira issue INFRA-18383 I wanted Git to prevent from modifications. 

Mistakenly `input` was wrong understood by me.
The value `input` makes changes to LF - don't want any change!

@vlsi Pls comment on this if you agree.

First the value was `true` then `input`. And both were wrong.
`false` is the right conservative value. Every project can change it inGit repo if modifications are needed.

vlsi:
@Tibor17 , `core.autocrlf=false` is wrong as well. 

For instance, https://svn.apache.org/repos/asf/maven/plugins/trunk/maven-checkstyle-plugin/ is configured to have `svn:eol-style=native`.

If you checkout that repository using default Windows machine, it produces text files with CRLF line endings. 
Did you just said "SVN has been corrupting Maven's sources for ages"?

Please have a look at https://www.apache.org/dev/svn-eol-style.txt
ASF recommends to use eol-style=native for SVN, and no-one states that "SVN corrupts sources".

`native` line endings is just the least surprise when developers use different platforms.

Tibor17:
@vlsi 
Again, I told you, the source must stay intact!!!
The PMD library chained their mind and their rules will no more rely on `native EOL`.
The Checkstyle plugin should be informed about PMD library and Checkstyle should adapt and be consistent.
The tendency of all is to be `neutral`.
Be neutral means intact sources.

vlsi:
@Tibor17 , feel free to configure your own machine in the way you want.

>  the source must stay intact

This claim has absolutely no justification.
On contrary, ASF suggests to use eol-style=native.

PMD and/or Checkstyle has nothing to do with autocrlf.

vlsi:
> The tendency of all is to be neutral

You don't get the point.
The tendency is to support building LF sources in Windows for the rare cases when someone downloads LF archive and tries to build it under Windows.

However, common settings for text editing is to use `native` line endings.

Tibor17:
@vlsi 
`LF` and `native` are wrong.
The point here is that Git installation done by INFRA should do this:
`$ git config --global core.autocrlf false`

Our responsibility in dev in project is to control our config.
We may or may not use
`$ git config --local core.autocrlf xxx`
or use `.gitattributes`.
BTW in my company we additionally used `.editorconfig`. We forced IDEA to LF this way, but that's upon a concrete project.

INFRA and their installations should stay neutral and should not touch the source code, whatever it means. Otherwise, if we find out any discrepancy between Jenkins Linux machine and personal Linuc machine, make sure that we report a bug against INFRA. This would happen if projects expect original sources in the build system!!!
or

vlsi:
@Tibor17 , I'm afraid you don't understand how those settings work, and it is sad you rush into altering global defaults instead of spending 5-10 minutes to understand the way it works.

It is really sad you claim "INFRA should do this" without bothering to justify your claims.

I have provided lots of examples (ASF recommendations, GitHub recommendations, SVN repository for your exact maven-checkstyle-plugin repo), and you seem to ignore those.


> INFRA and their installations should stay neutral and should not touch the source code, whatever it means 

Please study the difference between `core.autocrlf` and `.gitattributes`, then re-evaluate your thoughts.

> or use .gitattributes

`.gitattributes` cannot override `core.autocrlf`. Those settings complement each other.

Feel free to continue the discussion in https://issues.apache.org/jira/browse/INFRA-18763 or https://lists.apache.org/thread.html/8e3d4bde0a81ec7027a075b712ae6d1e68ed932e37e4995372d94169@%3Cbuilds.apache.org%3E

However I would assume you would provide at least one justification for your claims.

Regards.

PS. I apologize for the inconvenience to those who were accidentally subscribed for the notifications. I won't answer here anymore because it might notify more people that it was intended. I hope the discussion will continue in INFRA-18763 and/or builds@a.o

Tibor17:
@vlsi 
I need to speak with Gavin and Chris. These are two people in INFRA I know.

PR #1578, opened 2019-06-06
adding the full path to exclude

cthistle:
adding the full path to exclude


cthistle:
Just making this an OF and merging.

PR #1601, opened 2019-07-19
correct an IP matcher regex

gstein:
Ruby doesn't need escapes on the dots. Narrow the | operator. Switch from \d to [0-9] like the others.


FSchumacher:
Well, ruby needs to escape dots. If you don't escape the dots they are the usual regex dots -- meaning match any char.

On the other hand, matching ip addresses with regex is really hard and you are already using the ipaddr module. Have you tried to use it to match ip ranges like:
```ruby
case ipadd
  when IPAddr.new('140.11.11.0/24')
    "osuosl"
  when IPAddr.new('192.87.106.0/24')
    "sara"
...
```

PR #1605, opened 2019-07-23
Creates a PR for ofbiz-vm3

JacquesLeRoux:
This is only temporary IP change 
to test before ofbiz-vm2 is gone

There are also changes related to
INFRA-17523 "Add error page 503 for ofbiz demo sites"
Because I pulled after

I also wonder if we should not rather create an ofbiz-vm3.apache.org.yaml file ? At least at some point...


JacquesLeRoux:
I currently can't revert the change of the IP. But I prefer to create a new ofbiz-vm3.apache.org.yaml and keep ofbiz-vm2.apache.org.yaml as it was before this PR

JacquesLeRoux:
I did not know about https://github.com/apache/infrastructure-p6/blob/production/data/nodes/ofbiz-vm3.apache.org.yaml. 
Closing this PR

PR #1608, opened 2019-07-29
Alias for pmc-chairs

sebbASF:
This is needed to allow mbox to archive the pmc-chairs alias


gmcdonald:
What ticket is this in relation to? I have seen no request to have pmc-chairs alias archived?
(Also, maybe remove the 'ea' from the file as that role doesnt exist)

sebbASF:
This ticket is in relation to [1], fixing the pmc-alias.

AFAIK the alias has always been archived: see for example [2], and the current cronjob.

[1] https://lists.apache.org/thread.html/f268f9825fa70cf5c8d7794f1aa01605211b82dfcf40dfb4e2648431@%3Cprivate.infra.apache.org%3E
[2] https://mail-search.apache.org/members/private-arch/pmc-chairs/

PR #1617, opened 2019-08-04
Jmeter nightly new layout

FSchumacher:
With our migration to gradle we have changed the names of the artifacts. Therefore the index script has to be changed as well.

Modernize the script a bit while we are at it anyway.


sebbASF:
Note that cd can fail even if the directory exists.

Try the following:
$ mkdir /tmp/test
$ chmod -x /tmp/test
$ test -d /tmp/test && cd /tmp/test
I get:
-bash: cd: /tmp/test: Permission denied

Rather than the conditional block, it might be better to do something like

cd dir || { echo "Cannot cd to dir"; exit 1; }

PR #1618, opened 2019-08-04
Use IPAddr module instead of regex for ip calculations

FSchumacher:
The ipaddr module is used anyway in the customfact module,
so use it instead of brittle regex to check whether an
ip address is contained inside some network.


clambertus:
@rubys would you mind taking a look at this please? I'm not familiar enough with the module and the syntax to commit this without review. keep in mind this run on a very old version of ruby/puppet.

gmcdonald:
@rubys ping

gmcdonald:
Awesome thanks, I'll let @clambertus co-ordinate when to merge this

PR #1664, opened 2019-10-14
INFRA-19263: Threading of GitBox email messages

janhoy:
See https://issues.apache.org/jira/projects/INFRA/issues/INFRA-19263

Rough attempt at implementing threading support for PR notifications, so that all email sent by the script for the same project and same issue/PR id will be grouped in the same email thread by mail clients, as per RFC2822 - https://tools.ietf.org/html/rfc2822.html#section-3.6.4

For new issues/PRs only, the `Message-ID` header of the email is set as `<issue.$ISSUEID.$PROJECTNAME.gitbox@gitbox.apache.org>`, for all other emails the message id is auto generated as before. For followup issues/PRs, such as comments, the `In-Reply-To` and `References` headers are set to that exact stable id.

For obvious reasons the changes are not tested, so this is an input to INFRA which may need some testing and adjustment before ready.

### DIFF
--- a/modules/gitbox/files/cgi-bin/issues.cgi
+++ b/modules/gitbox/files/cgi-bin/issues.cgi
@@ -66,16 +66,20 @@ def getvalue(key):
     else:
         return None
 
-def sendEmail(rcpt, subject, message):
+
+def sendEmail(rcpt, subject, message, message_id=None, reply_to_id=None):
     if rcpt == 'dev@null':
         return
     sender = "GitBox <git@apache.org>"
+    if not message_id:
+        message_id = email.utils.make_msgid("gitbox")
+    reply_headers = "\nReferences: %s\nIn-Reply-To: %s" % (reply_to_id, reply_to_id) if reply_to_id else ""
     receivers = [rcpt]
     sub = email.header.Header(subject, 'utf-8').encode()
     msg = """From: %s
 To: %s
 Subject: %s
-Message-ID: %s
+Message-ID: %s%s
 Date: %s
 Content-Type: text/plain; charset=utf-8
 Content-Transfer-Encoding: 8bit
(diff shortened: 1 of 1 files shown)


sebbASF:
Are you sure that the initial Message-ID is going to be unique?

janhoy:
If GitHub sends exactly one “opened” callback per issue-id per project it should... Analyzing a bunch of callbacks could prove or disprove that assumption.

sebbASF:
There was a case in the past [1] when the PR numbers were re-used.

Maybe consider using a hash of the original PR subject or timestamp as well?
Obviously it would have to be something that can be derived from the follow-up messages.

I'm surprised that GitHub does not provide a unique id as part of the API.

Whilst it would be nice to be able to provide proper threading for PR mails, it's not essential, so I don't think it should not be at the expense of unique Message-Ids.

[1] https://issues.apache.org/jira/browse/INFRA-15973

janhoy:
Restarting numbering due to svn to git migration seems to me as a one-time situation that cannot re-occur?

If we want to guard against this in the future, the cgi script may keep a record of the current highest  github issue/pr per project, and if there is a “opened”-callback with a lower number we could generate a different I’d?

sebbASF:
As the JIRA says, the cause of that particular issue hopefully cannot recur. Whether there are other situations when it can occur, I don't know. e.g. can a second open event be sent?

However, I don't think it makes sense for the CGI script to keep a record of the highest PR number; that seems rather fragile.

Does the GitHub API really not provide a suitable unique ID?

Humbedooh:
AIUI, reopening an issue will send an 'open' event, not a 'reopen', so that could cause duplicate message-ids (we've seen this in the past)

Humbedooh:
I'm currently traveling, but I'll take a closer look and, if the PR works, merge it tomorrow morning :)

janhoy:
I setup a quick callback on one of my own repos and created an issue, commented, closed, repopened and closed again. Findings:
* There is ever only *one* `"action": "opened"` callback, reopening has `"action": "reopened"`
* The issue number has key `issue.number`, e.g. `2`. This should be unique in combination with project name.
* However I found a stable global IDs for the issue/PR, the `issue.node_id`

The node_id is part of all callbacks for this issue/PR, e.g. `"node_id": "MDExOlB1bGxSZXF1ZXN0MzI4NDE2NjA5"`. I'll update the PR to use it.

janhoy:
@Humbedooh Got any chance to test this?

PR #1678, opened 2019-11-09
feat: support GH protected branches in .asf.yaml

erisu:
# Feature Request

Support GitHub Protection Branch settings

## Motivation Behind Feature

It would be nice to be able to enable and disable on branches the protection settings without submitting an INFRA ticket.

### GitHub's Protection Settings:
* Require to pull request reviews before merging
  * Number of required review approvals
* Dismiss stale pull request approvals when new commits are pushed
* Require review from Code Owners
* Require status checks to pass before merging
  * Require branches to be up to date before merging
  * Context of a status check
* Require signed commits
* Include administrators
* Restrictions (Define what users, apps, and teams can push to the branch.)

## Feature Description

The goal of this PR is to implement as many of the defined settings above as possible.

**PR does not implement:**
* Context of a status check
* Include administrators
* Restrictions

APIv3 Repo Branches Documentation: https://developer.github.com/v3/repos/branches/

Example `.asf.yaml` configurations

```yaml
github:
  protected_branches:
    master:
      required_status_checks:
        strict: true
 
      required_pull_request_reviews:
        dismiss_stale_reviews: true
        require_code_owner_reviews: true
        required_approving_review_count: 3
 
      required_signatures: true
    branch_b:
      required_signatures: true
```

## Alternatives or Workarounds

Submitting an INFRA ticket in JIRA.

## Future Thoughts

* Support context of a status check
* Restrictions

## Test & Feedback

The remaining items.

- [x] Test setting and unsetting commit signing requirement
- [x] Test setting and unsetting `pull request reviews before merging`, `Number of required review approvals`, `review from Code Owners`, `status checks to pass before merging`, `branches to be up to date before merging`
- [ ] Test response `links` in `def getEnabledProtectedBranchList`. There are cases where if there are too many branches, the list will be paginated. I am trying to produce a list long enough to test that the linking works.
- [x] Any feedback from other...

### DIFF
--- a/modules/gitbox/files/asfgit/asfyaml.py
+++ b/modules/gitbox/files/asfgit/asfyaml.py
@@ -242,8 +242,8 @@ def getEnabledProtectedBranchList (GH_TOKEN, repo_name, url, isLast):
 
     return branchCollection
 
-def setProtectedBranch (GH_TOKEN, repo_name, branch, required_status_checks, required_pull_request_reviews):
-    REQ_URL = GH_BRANCH_PROTECTION_URL_TPL % (repo_name, branch, GH_TOKEN)
+def setProtectedBranch (GH_TOKEN, cfg, branch, required_status_checks, required_pull_request_reviews):
+    REQ_URL = GH_BRANCH_PROTECTION_URL_TPL % (cfg.repo_name, branch, GH_TOKEN)
     response = requests.put(REQ_URL, headers = {'Accept': GH_BRANCH_PROTECTION_URL_ACCEPT}, json = {
         'enforce_admins': None,
         'restrictions': None,
@@ -263,12 +263,12 @@ def setProtectedBranch (GH_TOKEN, repo_name, branch, required_status_checks, req
     title = "Protected Branches"
     message = "GitHub Protected Branches has been enabled on branch=%s" % (pb_branch)
     print(message)
-    notifiyPrivateMailingList(title, message)
+    notifiyPrivateMailingList(cfg, title, message)
 
     return response
 
-def removeProtectedBranch (GH_TOKEN, repo_name, branch):
-    REQ_URL = GH_BRANCH_PROTECTION_URL_TPL % (repo_name, branch, GH_TOKEN)
+def removeProtectedBranch (GH_TOKEN, cfg, branch):
+    REQ_URL = GH_BRANCH_PROTECTION_URL_TPL % (cfg.repo_name, branch, GH_TOKEN)
     response = requests.delete(REQ_URL)
     json = response.json()
 
@@ -283,12 +283,12 @@ def removeProtectedBranch (GH_TOKEN, repo_name, branch):
     title = "Protected Branches"
     message = "GitHub Protected Branches has been be removed from branch=%s" % (branch)
     print(message)
-    notifiyPrivateMailingList(title, message)
+    notifiyPrivateMailingList(cfg, title, message)
 
     return response
 
-def setProtectedBranchRequiredSignature (GH_TOKEN, repo_name, pb_branch, required_signatures):
-    REQ_URL = 'https://api.github.com/repos/apache/%s/branches/%s/protection/required_signatures?access_token=%s' % (repo_name, pb_branch, GH_TOKEN)
+def setProtectedBranchRequiredSignature (GH_TOKEN, cfg, pb_branch, required_signatures):
+    REQ_URL = 'https://api.github.com/repos/apache/%s/branches/%s/protection/required_signatures?access_token=%s' % (cfg.repo_name, pb_branch, GH_TOKEN)
     ACCEPT_HEADER = 'application/vnd.github.zzzax-preview+json'
 
     if type(required_signatures) is not bool:
@@ -313,7 +313,7 @@ def setProtectedBranchRequiredSignature (GH_TOKEN, repo_name, pb_branch, require
     title = "Protected Branches"
     message = "GitHub Protected Branches has set requires signature setting on branch '%s' to '%s'" % (pb_branch, required_signatures)
     print(message)
-    notifiyPrivateMailingList(title, message)
+    notifiyPrivateMailingList(cfg, title, message)
 
     return response
 
@@ -356,7 +356,12 @@ def formatProtectedBranchRequiredPullRequestReview(required_pull_request_reviews
 
     return required_pull_request_reviews
 
-def notifiyPrivateMailingList(title, body):
+def notifiyPrivateMailingList(cfg, title, body):
+    # infer project name
+    m = re.match(r"(?:incubator-)?([^-.]+)", cfg.repo_name)
+    pname = m.group(1)
+    pname = WSMAP.get(pname, pname)
+
     # Tell project what happened, on private@
     message = "The following changes were applied to %s by %s.\n\n%s\n\nWith regards,\nASF Infra.\n" % (cfg.repo_name, cfg.committer, body)
     subject = "%s for %s.git has been updated" % (title, cfg.repo_name)
@@ -458,7 +463,7 @@ def github(cfg, yml):
                 # GH API Calls to add/update
                 setProtectedBranch(
                     GH_TOKEN,
-                    cfg.repo_name,
+                    cfg,
                     pb_branch,
                     required_status_checks,
                     required_pull_request_reviews
@@ -466,7 +471,7 @@ def github(cfg, yml):
 
                 setProtectedBranchRequiredSignature(
                     GH_TOKEN,
-                    cfg.repo_name,
+                    cfg,
                     pb_branch,
                     required_signatures
                 )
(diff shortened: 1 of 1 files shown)


Humbedooh:
This is very cool! :)
One nit that I have is the try/except clauses that seem to drop the exception entirely. What we'd want, would probably be:
- in case of a http error, we'd want the JSON message from GitHub (so we can debug)
- in generic/other exceptions, we'd want the standard exception string (or just don't catch, but let the program bug out)

erisu:
Thank you for your review feedback. I will update the exception to include the JSON response from GH in the print out.

Humbedooh:
test comment, ignore :)

gmcdonald:
There is an increasing interest in this - see also https://issues.apache.org/jira/browse/INFRA-19907

erisu:
@Humbedooh Would you say in regards to the last commit, it would that cover your suggestion?

* I dropped the try catch.
* I raised an exception with a formatted message. It includes the original error message, status code, and that it is from a GitHub Request.
* I didn't handle any case if the request throws and error so the program should bomb out. 

```
if response.status_code != 200:
        raise Exception(
            "[GitHub] Request error with messaage: \"%s\". (status code: %s)" % (
                json.get("message"),
                response.status_code
            )
        )
```

I am wondering if a simple JSON message, such as "Bad Credentials", would be confusing so adding a format as seen above would be any helpful.

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -62,6 +62,132 @@ def pelican(cfg, yml):
...
+    return response
+
+def removeProtectedBranch (GH_TOKEN, repo_name, branch):
+    REQ_URL = GH_BRANCH_PROTECTION_URL_TPL % (repo_name, branch, GH_TOKEN)
+    response = requests.delete(REQ_URL)
+    json = response.json()
+
+    if response.status_code != 200:
+        raise Exception(
+            "[GitHub] Request error with messaage: \"%s\". (status code: %s)" % (
```
Humbedooh, line 115: typo, should be 'message'

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -62,6 +62,132 @@ def pelican(cfg, yml):
...
+    if required_signatures:
+        response = requests.post(REQ_URL, headers = {'Accept': ACCEPT_HEADER})
+    else:
+        response = requests.delete(REQ_URL, headers = {'Accept': ACCEPT_HEADER})
+
+    json = response.json()
+
+    if response.status_code != 200:
+        raise Exception(
+            "[GitHub] Request error with messaage: \"%s\". (status code: %s)" % (
```
Humbedooh, line 143: typo here as well.

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -113,7 +240,64 @@ def github(cfg, yml):
...
+                    cfg.repo_name,
+                    pb_branch,
+                    required_signatures
+                )
+
+        # Here is where the remaining branches as considered invalid/old branches with protection and will be removed.
+        # The cleanup process
+        for branch_to_disable_protection in enabledProtectedBranches:
+            removeProtectedBranch(GH_TOKEN, cfg.repo_name, branch_to_disable_protection)
+
```
Humbedooh, line 300: Should we include an email to private@$project.a.o when significant changes like these are being made to a repo?

erisu: I also think it might be a good idea to notify the project's private mailing list when a branch becomes protected or unprotected due to the nature of its change and affect.

Some people might ask if that email can be disabled. But, we can add an option later if there are many requests.

Do we have an example somewhere where the program emails private? I am not sure how I would add this.

Humbedooh: look in the notifications function just added recently

Humbedooh: https://github.com/apache/infrastructure-puppet/blob/deployment/modules/gitbox/files/asfgit/asfyaml.py#L490

gmcdonald:
What is left to do here @erisu @Humbedooh ?

erisu:
@gmcdonald I believe the last commit I pushed is what @Humbedooh is waiting for.

Need confirmation now.

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -221,6 +221,147 @@ def pelican(cfg, yml):
...
+    return required_pull_request_reviews
+
+def notifiyPrivateMailingList(title, body):
+    # Tell project what happened, on private@
+    message = "The following changes were applied to %s by %s.\n\n%s\n\nWith regards,\nASF Infra.\n" % (cfg.repo_name, cfg.committer, body)
+    subject = "%s for %s.git has been updated" % (title, cfg.repo_name)
+    asfpy.messaging.mail(
+        sender='GitBox <gitbox@apache.org>',
+        recipients=['private@%s.apache.org' % pname],
```
Humbedooh, line 365: Where is pname defined?

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -221,6 +221,147 @@ def pelican(cfg, yml):
...
+    elif required_pull_request_reviews is not None:
+        # If "required_pull_request_reviews" was not None but also it is not an object, we will force set it to None.
+        required_pull_request_reviews = None
+
+    return required_pull_request_reviews
+
+def notifiyPrivateMailingList(title, body):
+    # Tell project what happened, on private@
+    message = "The following changes were applied to %s by %s.\n\n%s\n\nWith regards,\nASF Infra.\n" % (cfg.repo_name, cfg.committer, body)
+    subject = "%s for %s.git has been updated" % (title, cfg.repo_name)
```
Humbedooh, line 362: Shouldn't cfg be passed to the mailer function?

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -273,7 +415,64 @@ def github(cfg, yml):
...
-        
+
+        # Fetches a collection of enabled protected branches.
+        # Branches will be removed from this collection if they exist in the ".asf.yaml" file with settings.
+        # Removing branches from this collection does not refer to removing protection settings.
+        # The reaming items are considered as old and invalid branches that have protection currently enabled.
+        # These branches will be stripped of its protection settings at the end.
+        enabledProtectedBranches = getEnabledProtectedBranchList(GH_TOKEN, repo_name, False, False)
+
+        if protected_branches:
```
Humbedooh, line 429: I'd throw in an assertion that protected_branches is a list of strings

modules/gitbox/files/asfgit/asfyaml.py
erisu: I thought `pname` was defined higher up but it seems it was not where I thought it was.
Since now `cfg` is passed down, I can create `pname`

modules/gitbox/files/asfgit/asfyaml.py
erisu: It is passed now.

erisu:
@Humbedooh I think I have now resolved the comments.. I did leave one question. Let me know if that was resolved now.

modules/gitbox/files/asfgit/asfyaml.py
Humbedooh: We are currently using python2.7 for this hook.

erisu: Ok, I have changed it to `basestring`

christ:
We can review this next team meeting, 06/11

casionone:
I want to know if it is possible to fuzzy match to the branch，such  as 
```
 protected_branches:
     dev-*
      required_status_checks:
        strict: true
        ......
```

reckart:
I would also like to know if wildcards are allowed in branch names - does anybody know?

reckart:
Related GitHub discussion: https://github.community/t/rest-api-v3-wildcard-branch-protection/13593/13

reckart:
It seems that via the REST API, it is not possible, but a person reports in the thread linked above that it should be possible via GraphQL: https://docs.github.com/en/graphql/reference/mutations#createbranchprotectionrule

shink:
Hi, @erisu, thank you for contributing to this awesome feature. But I have some problems.
In this [PR4257](https://github.com/apache/incubator-inlong/pull/4257), the following status checks must pass before merging.
But after this PR be merged, the problem is that not all status checks are triggered in every pull request. If the status checks are not triggered, the PR cannot be merged. This caused all PRs in the repository to not be merged, even the [PR4269](https://github.com/apache/incubator-inlong/pull/4269) that modify the `.asf.yaml` file. So, we cannot do any changes on master branch.
So, I reverted status checks in the [PR4272](https://github.com/apache/incubator-inlong/pull/4272). The goal is to pass all status checks and disable branch protection.
Fortunately, this PR can be merged, but after it be merged, the protected branch is still in effect. Finally, we had to ask for official help to unprotect branch settings. Please check out https://issues.apache.org/jira/browse/INFRA-23288?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

```yaml
contexts:
  - 'Build'
  - 'Unit Test'
  - 'Check license header'
  - 'Docker build and push'
  - 'Lint and test charts'
```

All in all, there are two problems:
1. All status checks must pass before merging. BUT if some status checks are not triggered, the PR will not be able to be merged.
2. After removing the protect branch setting, the setting is not updated.
Looking forward to your reply.

PR #1685, opened 2019-11-22
INFRA-19421

clambertus:
apply patch


gstein:
This is a good patch to apply. We don't want KEYS files on the mirrors. People downloading should only go to apache.org for the KEYS. We cannot guarantee mirrored KEYS files are authentic.

PR #1704, opened 2020-01-11
increase maxfiles for slapd

clambertus:
the process hangs when exceeding maxfile default 1024


clambertus:
self merging due to outage

PR #1711, opened 2020-01-16
INFRA-19665 Add an entry for asf948

martin-g:
It is a copy of asf946_ubuntu and it will be used as ARM64 BuildBot agent


gmcdonald:
no longer relevant

PR #1733, opened 2020-02-26
custom 404 for each docs version, see INFRA-19845

gmcdonald:
custom 404 for each docs version, see INFRA-19845


aljoscha:
@gmcdonald Thanks for looking into this! Should this now work or do we have do wait for the servers to pick this up? I'm still not being redirected to the 404 page on the docs, for example: https://ci.apache.org/projects/flink/flink-docs-master/getting-started/project-setup/java_api_quickstarto.html

gmcdonald:
There is still more to do, not as straight forward as it seemed. Will update on the Jira ticket when more is  done.

PR #1734, opened 2020-03-01
Migrated Jena from svnpubsub to gitpubsub

rlenferink:
The Jena [community voted](https://lists.apache.org/thread.html/ra90da10830d6575b75f2f43f08e2886e0fff898e6391092bd86dc3ae%40%3Cdev.jena.apache.org%3E) to migrate away from the Apache CMS and use git for serving its website.

After this `cms.a.o/jena` needs to be removed and the cms.conf needs to be edited.


rlenferink:
@gmcdonald for review

PR #1735, opened 2020-03-02
Dfoulks/custombuilder

dfoulks1:


modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -31,6 +31,69 @@ def jenkins(cfg, yml):
...
+    # Get notification list
+    pnotify = yml.get('notify', cfg.recips[0])
+
+    # Contact buildbot 2
+    bbusr, bbpwd = open("/x1/gitbox/auth/bb2.txt").read().strip().split(':', 1)
+    import requests
+    s = requests.Session()
+    s.get("https://ci2.apache.org/auth/login", auth= (bbusr, bbpwd))
+
+    if type(buildscript) is not 'string':
```
Humbedooh: should just be `str`, not `'string'`.
~~~
if type(buildscript) is not str:
    bork()
~~~

dfoulks1: good catch, done.

PR #1744, opened 2020-03-11
Notify pypubsub when public emails hit

Humbedooh:
This would notify pypubsub whenever a public (not private or restricted lists) hit our mail servers.
Intended use for now is for an overall ASF activity visualizer, and will only notify about standard open channels such as users@, dev@ etc.


sebbASF:
This seems risky to me.

What happens when the remote server does not reply immediately?

I think it would be better to write the data to a local file.
This can be watched by a separate process. 

Or use UDP notification.

PR #1757, opened 2020-04-16
Add notification scheme setter

Humbedooh:
To be documented :)
This adds notification scheme settings via .asf.yaml for repositories.
They are validated, then saved to a yaml file on disk.
To-Do: Use these settings...

### DIFF
--- a/modules/gitbox/files/asfgit/asfyaml.py
+++ b/modules/gitbox/files/asfgit/asfyaml.py
@@ -406,3 +406,74 @@ def publish(cfg, yml):
     except Exception as e:
         print(e)
         asfgit.log.exception()
+
+
+def notifications(cfg, yml):
+    """ Notification scheme setup """
+    valid_schemes = [
+        'commits',
+        'issues',
+        'pullrequests',
+        'issues_status',
+        'issues_comment',
+        'pullrequests_status',
+        'pullrequests_comment',
+    ]
+
+    # Get branch
+    ref = yml.get('refname', 'master').replace('refs/heads/', '')
+
+    # Ensure this is master, or bail
+    if ref != 'master' and ref != 'trunk':
+        print("[NOTICE] Notification scheme settings can only be applied to the master or trunk branch.")
+        return
+
+    # infer project name
+    m = re.match(r"(?:incubator-)?([^-.]+)", cfg.repo_name)
+    pname = m.group(1)
+    pname = WSMAP.get(pname, pname)
+
+    # Verify that we know all settings in the yaml
+    if type(yml) is not dict:
+        raise Exception("Notification schemes must be simple 'key: value' pairs!")
+
+    for k, v in yml.items():
+        if type(v) is not str:
+            raise Exception("Invalid value for setting '%s' - must be string value!" % k)
+        if k not in valid_schemes:
+            raise Exception("Invalid notification scheme '%s' detected, please remove it!" % k)
+        # Verify that all set schemes pass muster and point to $foo@$project.a.o
+        if not re.match(r"[-a-z0-9]+@[-a-z0-9]+(\.incubator)?\.apache\.org$" % pname, v)\
+            or not (
+                v.endswith('@%s.apache.org' % pname) or
+                v.endswith('@%s.incubator.apache.org' % pname)
+            ):
+            raise Exception("Invalid notification target '%s'. Must be a valid @%s.apache.org list!" % (v, pname))
+
+    # All seems kosher, update settings
+    scheme_path = os.path.join(cfg.repo_dir, 'notification_schemes.yaml')
+    print("Updating notification schemes for repository: ")
+    old_yml = {}
+    if os.path.exists(scheme_path):
+        old_yml = yaml.safe_load(open(scheme_path).read())
+    
+    # Figure out what changed since last
+    for key in valid_schemes:
+        if key not in old_yml and key in yml:
+            print("- adding new scheme (%s): %s" % (key, yml[key]))
+        elif key in old_yml and key not in yml:
+            print("- removing old scheme (%s) - was %s" % (key, old_yml[key]))
+        elif key in old_yml and key in yml:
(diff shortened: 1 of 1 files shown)


modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -406,3 +407,79 @@ def publish(cfg, yml):
...
+        'pullrequests_status',
+        'pullrequests_comment',
+    ]
+
+    # Get branch
+    ref = yml.get('refname', 'master').replace('refs/heads/', '')
+
+    # Ensure this is master, or bail
+    if ref != 'master' and ref != 'trunk':
+        print("[NOTICE] Notification scheme settings can only be applied to the master or trunk branch.")
```
gstein, line 429: the PMC won't actually see this notice, though. ?

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -406,3 +407,79 @@ def publish(cfg, yml):
...
+        return
+
+    # infer project name
+    m = re.match(r"(?:incubator-)?([^-.]+)", cfg.repo_name)
+    pname = m.group(1)
```
gstein, line 434: are all repo names proper? that regex will always match? ie. m != None

should we have something like `asfpy.util.extract_pmc(repo_name)` ? ... and that holds `WSMAP` and other related bits, for all ASF python code?

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -406,3 +407,79 @@ def publish(cfg, yml):
     except Exception as e:
         print(e)
         asfgit.log.exception()
+
+
+def notifications(cfg, yml):
+    """ Notification scheme setup """
+    valid_schemes = [
```
gstein, line 414: move this magic value to GLOBAL_CONST_VALUE

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -406,3 +407,79 @@ def publish(cfg, yml):
...
+        # Verify that all set schemes pass muster and point to $foo@$project.a.o
+        if not re.match(r"[-a-z0-9]+@[-a-z0-9]+(\.incubator)?\.apache\.org$", v)\
+            or not (
+                v.endswith('@%s.apache.org' % pname) or
+                v.endswith('@%s.incubator.apache.org' % pname)
+            ):
+            raise Exception("Invalid notification target '%s'. Must be a valid @%s.apache.org list!" % (v, pname))
+
+    # All seems kosher, update settings if need be
+    scheme_path = os.path.join(cfg.repo_dir, 'notification_schemes.yaml')
```
gstein, line 455: another magic value to move

should this be in the root dir on the repo, or in $repo/.git/ ?

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -406,3 +407,79 @@ def publish(cfg, yml):
...
+    m = re.match(r"(?:incubator-)?([^-.]+)", cfg.repo_name)
+    pname = m.group(1)
+    pname = WSMAP.get(pname, pname)
+
+    # Verify that we know all settings in the yaml
+    if type(yml) is not dict:
+        raise Exception("Notification schemes must be simple 'key: value' pairs!")
+
+    for k, v in yml.items():
+        if type(v) is not str:
```
gstein, line 442: `if not isinstance(v, str):`

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -406,3 +407,79 @@ def publish(cfg, yml):
...
+            raise Exception("Invalid value for setting '%s' - must be string value!" % k)
+        if k not in valid_schemes:
+            raise Exception("Invalid notification scheme '%s' detected, please remove it!" % k)
+        # Verify that all set schemes pass muster and point to $foo@$project.a.o
+        if not re.match(r"[-a-z0-9]+@[-a-z0-9]+(\.incubator)?\.apache\.org$", v)\
```
gstein, line 447: mebbe a global:
`RE_VALID_EMAIL = re.compile(...)`

not sure if we need that elsewhere? (ie. should such be moved to asfpy?)

modules/gitbox/files/asfgit/asfyaml.py
```diff
@@ -406,3 +407,79 @@ def publish(cfg, yml):
...
+    print("Updating notification schemes for repository: ")
+
+    # Figure out what changed since last
+    for key in valid_schemes:
+        if key not in old_yml and key in yml:
+            print("- adding new scheme (%s): %s" % (key, yml[key]))
+        elif key in old_yml and key not in yml:
+            print("- removing old scheme (%s) - was %s" % (key, old_yml[key]))
+        elif key in old_yml and key in yml and old_yml[key] != yml[key]:
+            print("- updating scheme %s: %s -> %s" % (key, old_yml[key], yml[key]))
```
gstein, line 473: do these print() calls end up in a log file somewhere? ... and/or should these be gathered into the email which is sent below?

modules/gitbox/files/asfgit/asfyaml.py
Humbedooh: Correct, only the pusher will see it, that is intended.

modules/gitbox/files/asfgit/asfyaml.py
Humbedooh: They end up in the git response when you push. The PMC gets an email with the updated settings, but I'm open to making that email have the same style.

modules/gitbox/files/asfgit/asfyaml.py
Humbedooh: We could do that, yes.
I think that's outside the scope of this particular PR, but something worth pursuing :)

Humbedooh: FWIW, it will always match. If it didn't, projects wouldn't have write access to them.

PR #1784, opened 2020-06-01
INFRA-20148 "Add RAT output files for release branches"

JacquesLeRoux:
Renames ofbiz.xml to ofbiz-trunk.xml
Adds ofbiz-next.xml and ofbiz-stable.xml


JacquesLeRoux:
Thanks Gavin, I have still an issue with Rat "next" build. I'll try to understand why...

PR #1838, opened 2020-12-31
INFRA-21250 rya-site project website: remove incubator from name

DLotts:
Not sure if this exactly what is needed, but it looks right.  I'm doing this to save you time so we can get this done before the new year!
Please pull ASAP so we can announce the release.  Whatever else needs to be done to update the website.
Happy new year!


clambertus:
this is not necessary if you are publishing via .asf.yaml, but no harm in merging it.

DLotts:
Site is updated!  Not sure why since you say this is not involved with the
yml configuration.
Thank you and happy new year!
 david.

PR #1839, opened 2021-01-06
INFRA-21065 Install libgdm1 to all Buildbot slaves

martin-g:


martin-g:
I hope `libgdm1` is available for these old versions of Ubuntu! I haven't verified this!

Apache Wicket uses only `bb_slave6_ubuntu` at the moment.

PR #1842, opened 2021-01-25
Allow all disk/repo-local references, lint.

Humbedooh:
There seems to be no reason to restrict local actions to the .github directory, and some projects are actively using local references outside that dir, so let's just allow it.


Humbedooh:
Outdated..

PR #1844, opened 2021-03-09
INFRA-20525 Add node config for wicket-vm2 (on Ubuntu 20.04)

martin-g:
Drop wicket-6.x Docker image - 6.x is not maintained anymore.
Use Tomcat 9.x instead of 8.x.

### DIFF
--- a/data/nodes/wicket-vm2.apache.org.yaml
+++ b/data/nodes/wicket-vm2.apache.org.yaml
@@ -1,0 +1,143 @@
+---
+classes:
+  - wicket_pvm_asf
+  - apache
+  - apache::mod::proxy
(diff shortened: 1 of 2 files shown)


gmcdonald:
Sorry for the mixup. With 20.04 nodes, we have ,moved to our infrastructure-p6 repository.
There, you will see there is already a yaml.

https://github.com/apache/infrastructure-p6/blob/production/data/nodes/wicket-vm2.apache.org.yaml

and the module is here

https://github.com/apache/infrastructure-p6/tree/production/modules/wicket_pvm_asf/manifests 

Please check those and see what changes might need doing

PR #1846, opened 2021-03-18
Add unsubscribe footer to emails

sebbASF:
Although email headers contain the information, many people don't know how to view them.


Humbedooh:
Sorry, had to revert this, as it turns out to add a bunch of unicode spaces between each character when pushing to github. I am unsure why...

PR #1853, opened 2021-03-20
Don't use the deprecated cgi.escape for Python 3

sebbASF:
https://github.com/git-multimail/git-multimail/commit/ddc9a1c


sebbASF:
ping?

Humbedooh:
We do not use python3 for multimail yet (a few months out), so no need to ping on this just yet.
I'll get it merged nonetheless.

PR #1856, opened 2021-03-27
Synch with p6

sebbASF:
Contents are out of sync with puppet 6

If the Tracking entries are no longer in use on puppet3, may I suggest that the section is emptied or removed?

[AFAICT other entries in the file are still used in P3]


sebbASF:
Accumulo edit is wrong

PR #1858, opened 2021-03-27
OF: Drop tracking entries to avoid accidents

sebbASF:


Humbedooh:
We'll be removing the entire module, see #1859

PR #1869, opened 2021-05-12
Exclude graduated podlings

sebbASF:
GitBox currently shows (Incubating) against podlings which have graduated


sebbASF:
Note: this only affects the index display: the titles and grouping

PR #1889, opened 2021-11-22
add ldap_pubsub role for authenticated access

clambertus:


Humbedooh (approved):
LGTM. Will need to see it in working condition, but I think this should do the trick.

